PR #299, opened 2021-02-06
Use main repository for releases.

nickygerritsen:
This will use the main repository for the releases.
My script that uploads ZIP's to the old releases is still running for a while but when it is done and we have merged this, we can get rid of the `build` repo.

I also fixed that the .gitlab-ci.yml will now get the version string from `version.properties` so we have only one place where it is stored.


nickygerritsen:
We can also get rid of the `SSH_PRIVATE_KEY` env var in GitLab after this is merged. `SSH_PRIVATE_KEY_WEBSITE` should stay though.

deboer-tim (approved):

PR #311, opened 2021-02-16
Support submission interface

deboer-tim:
First commit to support POST for submissions. RESTContestSource is updated to provide a method to perform a submit against a Contest API that supports it and the CDS' ContestRESTService supports POST by team or admin. The CDS acts as a simple proxy to the CCS.

Some details may need to be tweaked based on the final spec, but this has been tested with a local client able to submit to either the CDS or a temporary DOMjudge instance.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
+++ b/CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
@@ -1,10 +1,13 @@
 package org.icpc.tools.cds.service;
 
+import java.io.BufferedOutputStream;
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.StringTokenizer;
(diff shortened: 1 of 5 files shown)


CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
```diff
@@ -635,6 +641,149 @@ else if (!IContestObject.isSingleton(cType)) {
...
+		ContestSource source = cc.getContestSource();
+
+		if (cc.isTesting()) {
+			// when in test mode just accept submissions and return dummy id
+			sId = "test-" + obj.getString("team_id");
+		} else if (cc.getCCS() == null) {
+			response.sendError(HttpServletResponse.SC_BAD_REQUEST, "No CCS configured");
+			return;
+		} else if (!(source instanceof RESTContestSource)) {
+			response.sendError(HttpServletResponse.SC_BAD_REQUEST, "CCS does not support submissions via the CCS");
```
nickygerritsen, line 746: Strange line?

CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
```diff
@@ -635,6 +641,149 @@ else if (!IContestObject.isSingleton(cType)) {
...
+		}
+
+		ConfiguredContest cc = CDSConfig.getContest(segments[0]);
+		if (cc == null) {
+			response.sendError(HttpServletResponse.SC_NOT_FOUND, "Contest not found");
+			return;
+		}
+
+		String type = segments[1];
+		ContestType cType = IContestObject.getTypeByName(type);
```
nickygerritsen, line 674: On phone so can’t verify but what happens if you supply an unknown type? Does it throw or return null?

nickygerritsen (approved):

CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
deboer-tim: Ah, second instance should be CDS. This gives an error if the CCS is configured via XML event feed and hence doesn't support submissions via the CDS. Fixed via a commit.

CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
deboer-tim: If the client tries to post to a valid contest but to anything other than /submission they receive an error 400 ("Invalid type"). I've added another commit to improve the message to "POST can only be used for submissions".

PR #334, opened 2021-03-10
Clarification POST support

deboer-tim:
Initial support for posting clarifications and acting as a proxy to a CCS.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
+++ b/CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
@@ -672,8 +672,9 @@ public class ContestRESTService extends HttpServlet {
 
 		String type = segments[1];
 		ContestType cType = IContestObject.getTypeByName(type);
-		if (cType != ContestType.SUBMISSION) {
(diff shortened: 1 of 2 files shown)


CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
```diff
@@ -784,6 +793,77 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
 		pw.append("\"" + sId + "\"");
 	}
 
+	protected void handleClarification(HttpServletRequest request, HttpServletResponse response, ConfiguredContest cc,
+			JsonObject obj) throws IOException {
+		String id = obj.getString("id");
+		String time = obj.getString("time");
+		String fromTeamId = obj.getString("from_team_id");
```
nickygerritsen, line 800: No to_team_id support? (for admins)

deboer-tim: The JSON object is passed through, so admin request with to_team_id is just passed along. I realized I wasn't stopping teams from sending clars to another team though, so added a commit to block that.

nickygerritsen (approved):

PR #344, opened 2021-03-15
Determine base URL of API correctly.

nickygerritsen:
Instead of assuming the API of external CCS'es always starts with /api/, determine the real URL by using a regex on the CCS URL.
Kattis uses /clics-api/contests/xxxxx for a contest so it broke there if they use relative URL's.
Also use the root of the URL instead of the part before /api/ for absolute URL's, since I think that is what is meant in the spec.

@SuprDewd found this issue 😄.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/feed/RESTContestSource.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/feed/RESTContestSource.java
@@ -36,6 +36,8 @@ import java.util.Base64;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Locale;
+import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Tested with a test Kattis contest for NWERC20 btw

deboer-tim (commented):
Won't line 451 fail if the server is not on the default port #?

nickygerritsen:
https://stackoverflow.com/questions/2142910/whats-the-difference-between-uri-host-and-uri-authority

> Authority = Host Name + Port No

I actually copied it from the decompiled `url.toExternalForm` code.

deboer-tim (approved):
Ah, I was mis-reading.

PR #347, opened 2021-03-16
Add documentation on how to run everything inside IntelliJ IDEA.

nickygerritsen:
Since @SuprDewd asked for it and I almost lost this setup a few weeks ago I wrote a big document on how to get the tools running inside IntelliJ IDEA, including full debugging support in all tools (including the CDS!).

It's a lot of steps, but once you got it running, it should be fine.

I put it in a separate .md file to not clutter the main one. In the main one I already mentioned you can also use Eclipse, but I have never used it for that, so I leave that to @deboer-tim .

### DIFF
--- a/intellij-idea.md
+++ b/intellij-idea.md
@@ -1,0 +1,182 @@
+# Running the ICPC Tools inside IntelliJ IDEA
+
+You can use [IntelliJ IDEA](https://www.jetbrains.com/idea/) to develop and debug the ICPC Tools.
+Note that to work on the CDS, the Ultimate edition is required, since it uses Java EE.
+
+## Configuring general settings
+
+To get started, first open the whole repository in IDEA.
+
+Next, open File -> Project structure. Select a valid SDK under `Project SDK`. If none are available,
+click `Add SDK` and add one. Also select a directory under `Project compiler output` to put compilation results.
+
+## Configuring modules
+
+Next, go to Modules on the same window. Remove the `icpctools` module (or how it is called, depends on
+your directory name).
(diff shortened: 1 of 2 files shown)


nickygerritsen:
I wish we could easily reuse the IntelliJ project files, but I can't get it to work, it always contains some machine specific data.

intellij-idea.md
```diff
@@ -0,0 +1,182 @@
...
+  * Main class: `CoachView`
+  * Program arguments: `https://localhost:8443/api/contests/<id> public publ1c` where `<id>` is a configured contest ID
+* Floor generators:
+  * Module: `ContestUtil`
+  * Main class: `FloorGenerator<X>` with `<X>` the floor generator you want to use
+* Presentation admin:
+  * Module: `PresAdmin`
+  * Main class: `Admin`
+  * Program arguments: `https://localhost:8443/ admin adm1n`
+  * Note that the presentation admin requires a presentation client ZIP file inside a `present` directory in the same
```
nickygerritsen, line 164: @deboer-tim is this true? If I run the admin without any ZIP present I see no presentations so I assume it is.
I always thought we only used this ZIP for updating the client to a newer version.

deboer-tim: Yes, it is required - the admin pulls the list of presentations from the client zip sitting on the CDS. The CDS build does this automatically so we normally don't notice.

nickygerritsen: Ok. I couldn’t get IntelliJ to create a zip of another module during build so we just have to live with this for now

PR #351, opened 2021-03-18
Move contest card to client-side logic, separate out CDS-specific function

deboer-tim:
Further cleanup of server-side logic:
- Moved the CDS-specific ability for admins to add/remove/update any object from contest.js to it's own cds.js to keep contest.js with the spec. Used fixed base URL for the CDS.
- Changed contest (info) from JSP to javascript pages and renamed.
- Renamed awards.jsp to .html since it was already 'clean'.
- Moved calls to setContestURL() up front instead of waiting for page to load.


CDS/WebContent/js/cds.js
```diff
@@ -0,0 +1,69 @@
+var cds=(function() {
+	var contestURL;
+
+	var setContestId = function(contestId) {
+		contestURL = '/api/contests/' + contestId;
+		console.log("CDS URL: " + contestURL);
+	}
+	
+	var getURL = function(type) {
```
nickygerritsen, line 9: Not for now but this is overwritten by below function

nickygerritsen (approved):

PR #354, opened 2021-03-22
Upgrade AdminLTE theme

nickygerritsen:
The only change is the upgrade to AdminLTE 3.1, which is backwards compatible, so no scary changes.


deboer-tim (approved):
LGTM

PR #358, opened 2021-03-22
Add support for AdminLTE 3.1 dark mode

nickygerritsen:
Since @meisterT likes screenshots:
Screen Shot 2021-03-22 at 19 45 25
Screen Shot 2021-03-22 at 19 45 38
Screen Shot 2021-03-22 at 19 45 47

I also updated the website to have a dark layout, but there you can not switch; it will use the browser setting ([if your browser supports it](https://caniuse.com/?search=prefers-color-scheme)).

On the CDS it uses the browser based setting as default, but you can change it if you want. It will store it in the local settings, meaning it will persist for the current host indefinitely (unless you clear your local storage).

At first I had the theme menu as a submenu of the user menu, but that was not very logical I think. And we have more than enough space.


deboer-tim (approved):
LGTM

PR #360, opened 2021-03-23
Judgement and problem badges, consistent team display names, and UI f…

deboer-tim:
…ixes

- Consistent use of display_name for teams and "<id>: <display_name>" format.
- Changed submissions page to use judgement type badges instead of text judgement names.
- Clarifications now use problem badge.
- Fixed a problem caused by the last PR which made the scoreboard disappear for teams. :)
- Minor tweaks to column width, centering in a few places.

### DIFF
--- a/CDS/WebContent/WEB-INF/jsps/layout/head.jsp
+++ b/CDS/WebContent/WEB-INF/jsps/layout/head.jsp
@@ -145,7 +145,7 @@
 
               <ul class="nav nav-treeview">
                 <% for (int i = 0; i < menuPages.length; i++) 
-                   if ((i > 0 && i < 5) || Role.isAdmin(request)) { %>
+                   if ((i > 0 && i < 6) || Role.isAdmin(request)) { %>
                 <li class="nav-item">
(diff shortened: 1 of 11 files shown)


CDS/WebContent/WEB-INF/jsps/layout/head.jsp
```diff
@@ -145,7 +145,7 @@
 
               <ul class="nav nav-treeview">
                 <% for (int i = 0; i < menuPages.length; i++) 
-                   if ((i > 0 && i < 5) || Role.isAdmin(request)) { %>
+                   if ((i > 0 && i < 6) || Role.isAdmin(request)) { %>
```
nickygerritsen, line 148: Maybe at some point we should do this differently than hardcode the number of team menu items. Nor for now

nickygerritsen (approved):

CDS/WebContent/WEB-INF/jsps/layout/head.jsp
deboer-tim: Yes :)

PR #362, opened 2021-03-24
Country flags

deboer-tim:
Ok, I couldn't resist, country flags are shown next to the organization's country label. Flags courtesy of DOMjudge, originally from https://github.com/joielechong/iso-country-flags-svg-collection.


nickygerritsen (approved):

niemela:
These flags are not in the correct aspect ratios. 👎

PR #369, opened 2021-03-30
contest.js class

deboer-tim:
Change the contest.js to a proper class. This will allow some pages like the welcome page to connect to multiple contests at the same time and creates a better interface for clients. I also started using => to simplify the code.
This means we no longer support IE and currently 5% of the worldwide browsers according to https://caniuse.com/?search=class. This is unfortunate, but this is the cleanest and correct way to clean up the contest support for future and that number should be dropping.

### DIFF
--- a/CDS/WebContent/js/contest.js
+++ b/CDS/WebContent/js/contest.js
@@ -1,94 +1,121 @@
-var contest=(function() {
-	var info;
-	var state;
-	var organizations;
-	var groups;
-	var teams;
-	var languages;
-	var judgementTypes;
-	var problems;
-	var submissions;
-	var judgements;
-	var runs;
-	var clarifications;
-	var awards;
-	var startStatus;
-	var scoreboard;
+class Contest {
+	info;
+	state;
+	organizations;
+	groups;
+	teams;
+	languages;
+	judgementTypes;
+	problems;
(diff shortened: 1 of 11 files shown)


nickygerritsen:
If we would really care about that 5% we could use an ES6 transpiler. But imho we shouldn’t, I think no one who uses the CDS only has access to IE

CDS/WebContent/js/contest.js
```diff
@@ -1,284 +1,216 @@
...
 		    method: 'POST',
 		    headers: { "Accept": "application/json" },
 		    data: body,
-		    success: function(body) {
-		    	ok(body);
-		    },
-		    error: function(result) {
-			    fail(result);
-		    }
+		    success: ok(result),
```
nickygerritsen, line 204: I don't think this works? Or does it? Since it doesn't know what `result` is here.

I'd expect one of:
```suggestion
		    success: (result) => ok(result),
```
or
```suggestion
		    success: ok,
```

nickygerritsen (approved):

CDS/WebContent/js/contest.js
deboer-tim: As discussed, will fix next PR

PR #376, opened 2021-04-03
setStatus() when not modified

deboer-tim:
I noticed that Safari wasn't always loading org images. In the Web Inspector is looks like it's never getting a response, but on the server side you can see a setError(SC_NOT_MODIFIED) getting sent back. I didn't hook up a proxy to confirm, but this incorrectly includes the error page response in the body (like a 404) and I think that's what's messing up Safari. With this fix I've run several times and Safari is loading images correctly.


deboer-tim:
Oh, and including one file I thought had gone into the previous commit. :)

nickygerritsen (approved):

PR #409, opened 2021-06-05
Fix unicode character sanitizing

yang-er:
The character for full-width left quotation mark (`“`) cannot display correct in `head.jsp h1.m-0`.

Previously it was displayed as `&201c`. Then I found the `HttpHelper.sanitizeHTML(String s)` is not implemented correct before.

A probably fix is `&#8220;`, which doesn't require the hex converting.

BTW, there are many occurrence of unsanitized HTML. (e.g. `head.jsp li.has-treeview > a:first-child > p` is directly `<%= contest3.getName() != null ? contest3.getName() : "(unnamed contest)" %>`) Is the sanitization intended?


yang-er:
Before fix:
image

After fix:
image

deboer-tim (approved):

PR #410, opened 2021-06-05
Fix Timestamp parse in frontend

yang-er:
Currently, the timestamp string produced by CDS api is like `2014-06-25T11:22:05.034+01`.

However, in some browsers, this string cannot be parsed as Date, both in Firefox and Chrome.

firefox

chrome

If we extend the format be like `+01:00` it would work.

In Contest API standard, it said `timestamp` should be `yyyy-mm-ddThh:mm:ss(.uuu)?[+-]zz(:mm)?`. I think this change would be ok.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/feed/Timestamp.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/feed/Timestamp.java
@@ -11,18 +11,18 @@ public class Timestamp {
 	// parser for old spec times like: 1265335256.480
 	private static final Pattern OLD_TIME_PATTERN = Pattern.compile("([0-9]+)(\\.([0-9]{1,}))?");
 
-	// parser for times like: 2014-06-25T11:22:05.034+01
+	// parser for times like: 2014-06-25T11:22:05.034+01:00
 	private static final ThreadLocal<DateFormat> TIME_FORMAT = new ThreadLocal<DateFormat>() {
 		@Override
 		protected DateFormat initialValue() {
-			return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSX");
+			return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
 		}
 	};
-	// parser for times like: 2014-06-25T11:22:05+01
+	// parser for times like: 2014-06-25T11:22:05+01:00
 	private static final ThreadLocal<DateFormat> TIME_FORMAT2 = new ThreadLocal<DateFormat>() {
 		@Override
 		protected DateFormat initialValue() {
-			return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX");
+			return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX");
 		}
 	};
 


deboer-tim (commented):
On the parsing side, this fix is definitely required to meet the spec.

On the output side, I don't love this though. Clearly none of the CCS' I normally interact with (DOMjudge, Kattis, and PC^2) include the timezone minute in their output (or the CDS parsing would have had problems in the past) and I don't like outputting unnecessary information due to odd parsing behaviour in these browsers, especially since several CCS' would have this same behaviour.

Would you mind explaining what you're trying to do/tools you're using and why the output is directly in the browser? Thanks.

yang-er:
All my DOMjudge instances would output the minute section. This is the raw JSON fetched by curl.

```json
{"formal_name":"\u201c\u7ea2\u65d7\u676f\u201d\u7b2c\u5341\u516d\u5c4a\u9ed1\u9f99\u6c5f\u7701\u5927\u5b66\u751f\u7a0b\u5e8f\u8bbe\u8ba1\u7ade\u8d5b\uff08\u70ed\u8eab\u8d5b\uff09","penalty_time":20,"start_time":"2021-05-14T15:30:00+08:00","end_time":"2021-05-14T17:30:00+08:00","duration":"2:00:00.000","scoreboard_freeze_duration":"0:30:00.000","id":"3","external_id":null,"name":"\u201c\u7ea2\u65d7\u676f\u201d\u7b2c\u5341\u516d\u5c4a\u9ed1\u9f99\u6c5f\u7701\u5927\u5b66\u751f\u7a0b\u5e8f\u8bbe\u8ba1\u7ade\u8d5b\uff08\u70ed\u8eab\u8d5b\uff09","shortname":"hljcpc-warmup"}
```

Both `start_time` and `end_time` have the minute section. This is not related to configuration items in DOMjudge. I don't know whether this is related to the zone settings, but this behavior is showed on Ubuntu (both en-US and zh-CN).

---

The tool I'm using is CDS. There are many `Invalid Date Invalid Date` entries on CDS. For example,

image

The function `formatTimestamp` is from `icpctools/CDS/WebContent/js/ui.js`.

deboer-tim (approved):
Ok, I see where the problem is now. Timestamp.java does need to change in case the CDS is ever used where the timezone is not on the hour, but I confirmed it's ok otherwise. However, formatTimestamp() in ui.js definitely needs to change since it is reading a correct feed but not parsing correctly. Found the following in JavaScript reference: "Note: Parsing of date strings with the Date constructor is strongly discouraged due to browser differences and inconsistencies."

I'm going to accept this since it will make things work for now, but I may change the CDS timezone output back to +xx later and will definitely try to fix the Date parsing as the full fix.

Dup4:
In [List of UTC time offsets](https://en.wikipedia.org/wiki/List_of_UTC_time_offsets), we can find that there is a minute-level difference in the time difference between some areas and UTC.

Such as:

<p align="center">
  <a href="https://user-images.githubusercontent.com/38343778/120930040-2b0ab380-c71e-11eb-9d88-366fd83c8c97.png">
    
  </a>
</p>

And in the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#parameters)：

- The Date.parse() method parses a string representation of a date.
- The dateString representing a simplification of the ISO 8601 calendar date extended format. (Other formats may be used, but results are implementation-dependent.)

In the [ECMAScript specification](https://tc39.es/ecma262/#sec-date-time-string-format):

The stipulation is `HH:mm`.

<p align="center">
  <a href="https://user-images.githubusercontent.com/38343778/120931657-285f8c80-c725-11eb-8411-59a118bfd806.png">
    
  </a>
</p>

Although it is stipulated in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) that it is okay to keep only the `HH`:

<p align="center">
  <a href="https://user-images.githubusercontent.com/38343778/120931724-6eb4eb80-c725-11eb-9775-73863725acaa.png">
    
  </a>
</p>

So Contest API standard may follow the ISO 8601 specification.

But In [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339):

<p align="center">
  <a href="https://user-images.githubusercontent.com/38343778/120931764-9f952080-c725-11eb-9ef4-52aa51903df3.png">
    
  </a>
</p>

So, keep the format of `HH:mm`, is it a way with better compatibility?

PR #423, opened 2021-08-23
Update docs with host element

deboer-tim:
Updated the documentation to support the hosts element and related changes in order to stream multiple videos per team. While I was there I noticed some other minor issues and fixed them (e.g. reaction videos are now variable-length).


nickygerritsen:
Huh I see only changes to the prescontest readme?

PR #429, opened 2021-08-31
Remove pres admin dependency on contest source, model cleanup

deboer-tim:
Ok, I may as well be up front and admit that this is a bad idea. :) I still think the change is good and safe enough to make, but larger and more risky than you'd generally want right before a finals...

While fixing the presentation admin I realized that as much as I could make the REST contest source not force a connection to a particular contest, that's kind of missing the point (not what the class is meant for). So, I separated out the CDS update utility and made the command line argument parser and basic client allow use without a contest source.

Before I got to this point I was trying to clean up the contest sources and the myriad ways that you could connect to or pull data for a contest. This change simplifies and makes these much more consistent: there are just two classes (disk and REST) and they have fewer, more obvious constructors. See notes in ContestSource for details.

Other notes:
- Added support for reading from an event feed to the Disk source, and deleted the separate EventFeed (file) source. There is now just one class that can read from disk.
- CDS contest source is always a DiskContestSource, so I exposed the type and removed a few if statements.
- Tools like the award generator and event feed util (maybe interesting) and balloon printer (less interesting) could now load from a contest folder (CAF) instead of just a standalone feed.


nickygerritsen (approved):
Jeebz Tim. Not gonna check all this code on a phone. I believe you

PR #436, opened 2021-09-15
Emoji (take II)

deboer-tim:
A recent PR added emoji support via a series of fallback fonts. While this worked, it had a number of limitations:
- several fonts & the complexity of that
- couldn't handle zero-width spacers and other unique characters
- no fill or colours

I tried writing some additional code to handle zero-width chars, but if you put two emojis on top of each other (without color or fill) it becomes unreadable. The only way to do this well (until the language and Fonts add native support) is to use a png for each emoji. The ~3000 pngs included here are from the same Twitter colour emoji font we were using.

I'm sure there is a presentation or two that I missed adding support to, and there are some other code patterns that could probably be simplified or helpers provided for (e.g. text alignment options, or putting a real image beside text). OTOH this is significantly better than what we had before and covers the majority of cases, with full colour emojis and lots of special case/combined emojis.


deboer-tim:
Almost forgot: Screen Shot 2021-09-15 at 2 11 06 PM

Issue #402, opened 2021-05-05
Support Unicode / Emoji in Team Names


mpsijm:
Suggestion: better support for Unicode characters or emoji in team names in the Resolver (and probably in other tools as well).

Take for example these two team names from [BAPC 2020](https://2020.bapc.eu/scoreboard):
<details>
<summary>Screenshots</summary>

![image](https://user-images.githubusercontent.com/9739541/117210660-48cfba80-adf8-11eb-85e8-c7279da45e09.png)

The default font (Helvetica) which is shipped with ICPCtools does not support many of these characters:
![image](https://user-images.githubusercontent.com/9739541/117210763-69981000-adf8-11eb-85da-82c164fe8e80.png)

This already improves when switching to a different font, I took DejuVu Sans as an example:
![image](https://user-images.githubusercontent.com/9739541/117210873-8e8c8300-adf8-11eb-99c5-b9819458e434.png)

However, as far as I know, there does not exist a font that supports both "regular" letters and symbols and (color) emoji at the same time. For example, the [twemoji](https://github.com/eosrei/twemoji-color-font) font _only_ supports emoji but no letters:
![image](https://user-images.githubusercontent.com/9739541/117211437-4457d180-adf9-11eb-8635-25623ca0f366.png)
</details>

I see two challenges here:
1. Use two fonts at once (one main font and a fallback font). This can be solved by:
   - Merging multiple fonts together to have it support both "regular" symbols (main) and emoji (fallback). I've tried [merging fonts](https://superuser.com/a/1594984) with [FontForge](https://fontforge.org/), but without success.
   - [generating an `AttributedString`](https://stackoverflow.com/a/9482676) and printing that. This would require changes at all locations in the code where a team name is printed, though.
2. In the last screenshot, the emoji are rendered in black-and-white, but it would be better if they are in color. This [seems to be possible](https://stackoverflow.com/a/43950434), but I have not tested this yet. 

I just wanted to ventilate my thoughts before trying to code this up and create a PR. Any comments or hints are appreciated :slightly_smiling_face:

deboer-tim:
Thanks for the bug report. I've been aware of this for years but haven't found a solution nor copied the bug report from the bugzilla system we used to use. I last looked at this a couple years ago (and several times before that) and couldn't find any reliable way to do it. I still have a changeset with OpenSansEmoji.ttf as a fallback font that only partially worked... I was trying to check each character in the string and switch fonts when necessary.

[...]

PR #456, opened 2021-09-24
TextHelper: give emoji PNGs priority over font glyphs

mpsijm:
Based on the follow-up discussion in #402:

This PR simplifies the rendering of emoji. Previously, the emoji PNGs were only consulted when the main font did not contain a glyph for a certain Unicode code point. Now, the code points of a string are first scanned to see if they contain sequences that can be printed using Twemoji's PNGs, which includes sequences that contain the Zero-Width Joiner.

Additionally, this PR contains some general refactorings to the TextHelper class, and a way to override the `ICPC_FONT` environment variable using a local font file.


mpsijm:
Obsolete with the changes from #455 (again, thanks a lot for supporting emoji! :tada:)

PR #476, opened 2021-10-01
Sync time immediately

deboer-tim:
Add a method to schedule time syncing immediately, and call it whenever we sync the time.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/presentations/Client.java
+++ b/CDS/src/org/icpc/tools/cds/presentations/Client.java
@@ -226,7 +226,14 @@ public class Client {
 		});
 	}
 
-	protected long handlePing() throws IOException {
+	/**
+	 * Handle a ping. Returns true if we've decided to sync time and need to send an event
+	 * immediately, false otherwise.
+	 *
+	 * @return
+	 * @throws IOException
+	 */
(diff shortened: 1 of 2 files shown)


zalenski (approved):

CDS/src/org/icpc/tools/cds/presentations/Client.java
```diff
@@ -258,27 +265,27 @@ protected long handlePing() throws IOException {
...
-			// very inconsistent response times, ping once more in 750ms
+			// very inconsistent response times, ping at least once more
 			writePing();
-			return 750;
+			return false;
 		}
 
-		// get rid of 'worst' time outlier and determine new average time sync
+		// we've had at least 3 pings, get rid of 'worst' time outlier and send average time sync
 		writeTime((total - max) / (count - 1));
```
zalenski, line 287: ideally, the client currentTimeMillis should be communicated, so cds could send a time delta value to the client based on the shortest pings. with the current writeTime, it is very sensitive to the delay of _that_ call.

zalenski: (not a big problems as long as pings and requests take milliseconds)

PR #492, opened 2021-10-02
Add CORS to stream endpoints.

nickygerritsen:
As requested by @SteinNorheim


nickygerritsen:
Couldn't get the <cors> server.xml thing to work and saw we already did it this way for the API so copied it to the stream service.

deboer-tim (approved):

PR #536, opened 2021-10-11
Fix reading event feeds from file

mpsijm:
I was not able to run the Awards tool or Resolver tool from an `event-feed.json` file. Instead, I had to pass the path to the directory containing this file as argument to these tools. It turns out that, when passing the path to an `event-feed.json` file, the tools tried to parse this path as a `RESTContestSource`, while I guess this should be a `DiskContestSource`. This bug seems to be present since e4bcfd056a25db180a5d214475aefb2ed24d47d9.

While I was at it, I also updated the documentation of the Awards tool a bit, because I had to look in the source code to see what some of the arguments actually did :stuck_out_tongue:


mpsijm:
Awesome, thanks for upgrading my fix to something more sustainable in #541! :smile: 

I've rebased away the feed commit, and left in the commit that fixes documentation in Awards.java :slightly_smiling_face:

PR #577, opened 2021-11-20
Allow team_ids and problem_ids to be null for commentary.

nickygerritsen:
The spec says problem_ids and team_ids can be null but we didn't support it properly. Now we do.


deboer-tim (approved):
If the commentary has a team or problem array and then it is changed to null via addImpl() the value wouldn't take effect. Fairly unlikely in practice, but maybe just add an 'else teamIds = null;'?

nickygerritsen:
> If the commentary has a team or problem array and then it is changed to null via addImpl() the value wouldn't take effect. Fairly unlikely in practice, but maybe just add an 'else teamIds = null;'?

Sounds reasonable. Will do tomorrow

PR #578, opened 2021-11-20
Allow two configured contests with the same CAF but another CCS URL.

nickygerritsen:
So NWERC uses the same CAF location for both Kattis and DOMjudge, but (of course) used different CCS URL's. The cache directories where the same so we had a event feed cache file again that was corrupted. This uses the CCS URL (if present) in the cache directory name, preventing this issue.


deboer-tim (approved):
I feel like there is still some discussion required on how this should work - e.g. should it be building a single folder structure that covers the data from both; should the feed be in the CAF structure, etc. - but this is definitely a step to avoid failures before that.

nickygerritsen:
Yeah I was also not sure. But this is clearly better

PR #584, opened 2022-03-09
When current list is null or empty, use local list.

nickygerritsen:
My CDS gave an error if I had JSON's without logo's but then had logo's on disk. This fixes it, but I'll leave it to you to decide if this makes sense.


deboer-tim (approved):
I'm being picky, but maybe move this above the variable assignment and make it a (opposite) copy of lines 846/847?

PR #585, opened 2022-03-09
Add option to use team intro presentation for organizations.

nickygerritsen:
cc @niemela , who suggested this during NWERC.

This uses the same team intro presentation, but then uses it for organizations. Useful if you have multiple teams per org.

I tried it in the SWERC dataset. It *does* work, but since uni's are still pretty close in these contests, it doesn't look great. Maybe we later should improve that.

Note I had to call `setup()` again to re-set up the data. Changing the property while the presentation is running makes it flicker for a few ms, but I'm not sure if we can fix that.


deboer-tim (approved):
LGTM

PR #603, opened 2022-03-30
Add title to /team-member and add temporary /persons endpoint

deboer-tim:
Made the following changes:
- Make /api respond to requests in addition to /api/. Technically this isn't required by the spec, but I think it's better to do both.
- Add title to team member, the one property that's missing from person.
- Make /persons an alias to /team-members. This is a temporary hack that would be removed when team members is renamed.

The last change means that the /persons endpoint can be supported as per the spec *if read from disk* (or from a CCS that supports team-members and you don't need the new properties) and *if you don't care about the event-feed*. It's temporary, it's ugly, but it's a really cheap way to get a working /persons endpoint for now. FWIW I thought about implementing a full alias or putting both types on the feed, but that seemed like overkill.


nickygerritsen (approved):
I think the back is decent for now

PR #610, opened 2022-04-08
Users -> Accounts, HTTPS, and other things

deboer-tim:
This change started by trying to replace users.xml with an accounts.json. Switching to accounts allows us to easily copy/paste between the CDS auth and contests and share account types. In order to do this though:
- I had to rewrite the CDS' auth later and create an identity store, update web.xml and Liberty config, etc.
- Had to change the names of the old roles, e.g. "blue" -> "staff" and "trusted" -> "analyst".
- We had CDS (global) accounts and contest-level accounts before, but now it's more obvious and these are almost the same thing. Tried to change the code where appropriate to use the right one of these. e.g. someone could be analyst while logging into the CDS, a team in one contest, and staff in a second. When looking at contest data, you typically want to base visibility on your contest account.

Since the changes above took a while, several other changes crept in, and it's easier to include here vs PR independently (also, I'm lazy):
- I had already planned to simplify the ConfiguredContest logic for who-has-access-to-what, and ended up creating a bunch of classes in .contest.model.internal.account. I put these in the contest model because they follow the spec accounts, and I expect them to be reusable anywhere else you want to know what a particular type of account can see.
- Changed 'Login' to 'Sign in'.
- Updated sample config.
- Removed old Role and BasicAuthFilter code.
- Started implementing /access.
- HTTPS is now used for everything.

TODO:
There are a few places that are clearly incomplete and I plan to address in a following PR. Normally I would hold until these are fixed, but at this point I think it's more important to get the rest of the changes building.
- Several topics need to be discussed on the next CLICS call, including sharing of accounts.json, global vs contest accounts, extending the account types, etc. These could impact this code and lead to further change.
- Synchronization: if someone logs into a contest while events are coming in, one could be lost.
- You can't do basic auth via a web browser, but if you managed to do this while testing :) there's no way to log out. Have an idea, need to try it.
- The switch to HTTPS could tank performance of video streams, need to test or back the change out just for video.


CDS/src/org/icpc/tools/cds/presentations/PresentationWebSocket.java
```diff
@@ -105,7 +104,7 @@ public void onOpen(Session session, EndpointConfig config) {
 			}
 		} else if (!"any".equals(role)) {
 			boolean isBlue = info.isBlue;
-			if (((Role.PRES_ADMIN.equals(role)) && !isAdmin) || (Role.BLUE.equals(role) && !(isAdmin || isBlue))) {
+			if ((("presAdmin".equals(role)) && !isAdmin) || ("staff".equals(role) && !(isAdmin || isBlue))) {
```
nickygerritsen, line 107: What's the reason we are now hardcoding the role names? Shouldn't we still have a constant somewhere to make it easier to change it?

CDS/src/org/icpc/tools/cds/CDSAuth.java
```diff
@@ -0,0 +1,146 @@
...
+import org.icpc.tools.contest.Trace;
+import org.icpc.tools.contest.model.IAccount;
+
+@ApplicationScoped
+public class CDSAuth implements HttpAuthenticationMechanism {
+	private static final String AUTHORIZATION_HEADER = "Authorization";
+	private static final String BASIC_PREFIX = "Basic ";
+	private static final String RESULT = "result";
+	private static final String ACCOUNT = "account";
+
```
nickygerritsen, line 29: So perhaps we should move those constants here?

CDS/src/org/icpc/tools/cds/CDSAuth.java
```diff
@@ -0,0 +1,146 @@
...
+		// check for host-based auto-login
+		if (request.getRemoteUser() == null) {
+			CDSConfig config = CDSConfig.getInstance();
+			username = config.getUserFromHost(request.getRemoteHost());
+			if (username == null)
+				username = config.getUserFromHost(request.getRemoteAddr());
+
+			if (username != null) {
+				password = config.getUserPassword(username);
```
nickygerritsen, line 69: Could we (if we wanted) now drop the requirement that a password is required in the JSON if you do IP auth? I'm not too familiar with the Java auth stuff, but does the UsernamePasswordCredential below need the password? I guess it does?

CDS/src/org/icpc/tools/cds/CDSConfig.java
```diff
@@ -337,13 +347,18 @@ protected String getUserFromTeamId(String teamId) {
 		if (teamId == null || teamId.isEmpty())
 			return null;
 
-		for (ConfiguredContest cc : contests) {
+		for (IAccount account : userAccounts) {
```
nickygerritsen, line 350: The comment above this method is not really correct anymore I think

CDS/src/org/icpc/tools/cds/CDSIdentityStore.java
```diff
@@ -0,0 +1,26 @@
...
+import javax.enterprise.context.ApplicationScoped;
+import javax.security.enterprise.credential.UsernamePasswordCredential;
+import javax.security.enterprise.identitystore.CredentialValidationResult;
+import javax.security.enterprise.identitystore.IdentityStore;
+
+import org.icpc.tools.contest.model.IAccount;
+
+@ApplicationScoped
+public class CDSIdentityStore implements IdentityStore {
+	public CredentialValidationResult validate(UsernamePasswordCredential userCredential) {
```
nickygerritsen, line 16: Ah so maybe we *could* do this for IP auth, if we add some magic here. Although I'm not sure if we really want it

CDS/WebContent/WEB-INF/jsps/layout/head.jsp
```diff
@@ -24,6 +24,13 @@
 <script src="${pageContext.request.contextPath}/js/bootstrap.bundle.min.js"></script>
 <script src="${pageContext.request.contextPath}/js/adminlte.min.js"></script>
 <script src="${pageContext.request.contextPath}/js/theme.js"></script>
+<script>
+function logout() {
```
nickygerritsen, line 28: What does this magic do?

ContestModel/src/org/icpc/tools/contest/model/internal/account/PublicContest.java
```diff
@@ -0,0 +1,185 @@
...
+	public void add(IContestObject obj) {
+		IContestObject.ContestType cType = obj.getType();
+
+		switch (cType) {
+			// all of these are fully public
+			case CONTEST:
+			case JUDGEMENT_TYPE:
+			case MAP_INFO:
+			case START_STATUS:
+			case AWARD:
```
nickygerritsen, line 54: Are awards not updated during the freeze? The spec says the freeze should be taken into account. Imagine a CCS updating awards during the freeze, that would not be OK if the public sees that, right?

ContestModel/src/org/icpc/tools/contest/model/internal/account/PublicContest.java
```diff
@@ -0,0 +1,185 @@
...
+					int freezeTime = getDuration() - getFreezeDuration();
+					if (time >= freezeTime)
+						return;
+				}
+
+				// don't show judgements for hidden teams
+				if (isJudgementHidden(j))
+					return;
+
+				super.add(j);
```
nickygerritsen, line 146: The spec says this is fine, but do we actually want to show the public the exact verdict for submissions before the freeze? Not for this PR, but didn't we talk about it during a CLICS call?

ContestModel/src/org/icpc/tools/contest/model/internal/account/BalloonContest.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.icpc.tools.contest.model.IAccount;
+import org.icpc.tools.contest.model.IContestObject;
+import org.icpc.tools.contest.model.IContestObject.ContestType;
+import org.icpc.tools.contest.model.IJudgement;
+import org.icpc.tools.contest.model.ISubmission;
+
+public class BalloonContest extends PublicContest {
```
nickygerritsen, line 12: Mayba also add a comment to this class as to what it filters on?

nickygerritsen (commented):
Some smaller comments but I like it!

CDS/src/org/icpc/tools/cds/presentations/PresentationWebSocket.java
deboer-tim: Fixed in new commit. I was hesitating before b/c some of them aren't 'standard', but added these to IAccount since they are account types.

CDS/src/org/icpc/tools/cds/CDSAuth.java
deboer-tim: As above, went with IAccount since these are account types.

CDS/src/org/icpc/tools/cds/CDSAuth.java
deboer-tim: I *think* we could do that now since we control both identity store and auth, but decided to leave investigating that until later.

CDS/src/org/icpc/tools/cds/CDSConfig.java
deboer-tim: Doh. Fixed javadoc and removed commented out code (an earlier attempt at logging in based on every configured contest...) in new commit.

CDS/src/org/icpc/tools/cds/CDSIdentityStore.java
deboer-tim: Agreed. One implication is that you wouldn't be able to log into that account from any other machine.

CDS/WebContent/WEB-INF/jsps/layout/head.jsp
deboer-tim: Nothing, at the moment. :) A hack I saw online to log out of Basic auth in a web browser is to try logging in with an incorrect user in the background, was starting to try that and haven't finished.

ContestModel/src/org/icpc/tools/contest/model/internal/account/PublicContest.java
deboer-tim: If you take a look at the code before, awards weren't even output to roles below blue/staff. Maybe I should just hide them again now and do the full logic via another PR? Or ok to defer this fix?

ContestModel/src/org/icpc/tools/contest/model/internal/account/PublicContest.java
deboer-tim: Agreed, I thought there was a discussion about a simplified set of judgement types for public. I can't find anything in the spec now though, will add to the agenda.

ContestModel/src/org/icpc/tools/contest/model/internal/account/BalloonContest.java
deboer-tim: Added a description, and noticed a simpler way to determine num solved. New commit pushed.

PR #622, opened 2022-04-14
Expose full bounds to presentations for floor map.

nickygerritsen:
Also expose balloons to draw paths during testing.


deboer-tim (approved):
I remembered the original reason for this was to avoid scaling for (hidden) aisles onscreen. This is a good fix for now, maybe I can change the boolean meaning later.

PR #626, opened 2022-04-20
Replace the user configuration with accounts.json

### LINKED ISSUE #625: Docker user accounts are incorrect in CDS
For the docker image of cds version 2.5.737 the configuration of the method of user accounts is outdated. It still uses the old `users.xml`, while the cds looks in `accounts.json` as described in the manual. Setting the password through environment variables has currently no effect, since they get inserted in the wrong file.

A work around is to link your own `accounts.json` from the local file system in the docker container to `/opt/wlp/usr/servers/cds/config/accounts.json`.

verwoerd:
Should fix #625


nickygerritsen:
Superseeded by #632.

Issue #628, opened 2022-04-20
Per-team video configuration does not work


MarcoUCM:
When cdsConfig.xml contains a video element for an individual team such as

```
<contest ...>
	 <video
		  id="team01"
		  desktop="http://192.168.10.34:9999"
		  desktopMode="lazy"/>
</contest>
```

 the same configuration is applied to every team (its mapped to different streams in the VideoAggregator).

deboer-tim:
Hi Marco,
Thank you for the bug report and PR! As you know I changed the fix slightly, but this should now be fixed. Please feel free to reopen if you have issues.

PR #629, opened 2022-04-20
Fix per-team video configuration (cdsConfig.xml)

MarcoUCM:
Should fix issue #628


deboer-tim:
This would fix the issue, but wouldn't cache the result for next time. Likely not a big issue, but I've submitted PR 630 that includes this change as well as some other caching/cleanup.

PR #638, opened 2022-04-27
Animated 'no presentation'

deboer-tim:
Two minor requests from Moscow: make the 'no presentation' screen animated and add the IP address. The logical step (and simpler code overall) was to just use a presentation for this screen too. I've created one, and called it NoPresentation (was tempted to use ThisIsNotAPresentation).

I moved the ICPC Tools presentation to the core package since it shares a lot of code & images with the new one.

Nits: changed folder to warning icon when exiting admin, and fixed logosOff property. Also fixed a race condition where the presentations weren't the right size if they loaded too quickly starting up.


nickygerritsen (approved):
This definitely needs a demo video 🤣

deboer-tim:
It's the same as before, but the ICPC Tools logo is animated. Or conversely, it looks just like the ICPC Tools presentation, but smaller and with "no presentation" at the bottom. 😛

PR #649, opened 2022-05-27
Improve relative URL detection

deboer-tim:
I added 'if (!href.startsWith("http"))' a few weeks ago to avoid the case where a local event feed included file reference urls and this was causing it to try to download weirdly named files, causing lots of errors in logs. Unfortunately, it was causing relative URLs to be skipped as well.

The correct fix is in getResolvedURL(), but along the way I noticed that it was only being used in createConnection() from one caller, so inlined that and removed the boolean argument, simplifying other calls to createConnection().


nickygerritsen (approved):
Lol we in DJ had a similar issue where we thought `/clics-api/xxx` was relative but it clearly isn't.

PR #671, opened 2022-07-06
FIX: encodeValue(string) didn't quote the string

MarcoUCM:
The JSON messages sent to presenter clients are not correct in the contest.ids field (some values are omitted):

```
{"type":"clients","clients":[
   {"name":"...","uid":"...","width":1920,"height":1080,
   "contest.ids":[TestContest/],
   "version":"2.5.795","client.type":"presentation",
   "displays":[...]}
]}
```
The JSON encoder do not quote the id in the array (note the TestContest/ string).


deboer-tim:
Good catch, thanks for the PR.

PR #674, opened 2022-08-10
Use curl to upload assets so we can retry

nickygerritsen:
Thanks @troy2914 for the commands, this uses cURL so we can retry failed uploads.


deboer-tim (approved):
LGTM

PR #676, opened 2022-09-20
Fix event feed parsing for new format for contest type.

nickygerritsen:
@deboer-tim this seems to be the easiest way to expose `contests` as `contest` in the new feed format AND still have all tools work internally; I moved the check for 'incoming' messages to the generic spot and that seems to work.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/IContestObject.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/IContestObject.java
@@ -37,6 +37,10 @@ public interface IContestObject {
 	}
 
 	static ContestType getTypeByName(String typeName) {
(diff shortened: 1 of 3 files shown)


deboer-tim:
This one has some other odd consequences (e.g. you could PUT a contest), but since those are new behaviours and wouldn't break anything related to the spec, I'd be fine to approve for now and fix those later.

nickygerritsen:
> This one has some other odd consequences (e.g. you could PUT a contest), but since those are new behaviours and wouldn't break anything related to the spec, I'd be fine to approve for now and fix those later.

Ah yeah that makes sense. I agree that merging it is better than it is now and then we fix it later.

PR #678, opened 2022-09-25
CLICS Specs are currently hosted at github.io

vmcj:
There are more mentions of the site at baylor, will that server be recovered?


nickygerritsen:
Nope, Baylor won’t return probably

nickygerritsen (approved):

PR #705, opened 2022-11-01
CMS config generation fixes

deboer-tim:
- Fix person to output team_ids property.
- Image generation fix for tile (160x160px) images.
- Sort persons.json by id (sorting by team was causing more diffs).
- TSV importer fix for persons with multiple teams.

### DIFF
--- a/ContestUtil/src/org/icpc/tools/contest/util/cms/JsonToTSVConverter.java
+++ b/ContestUtil/src/org/icpc/tools/contest/util/cms/JsonToTSVConverter.java
@@ -109,7 +109,7 @@ public class JsonToTSVConverter {
 					return false;
 				}
 			});
-			
+
 			System.out.println("Loading " + files.length + " institutions");
 			for (File f : files) {
 				readInstitution(f);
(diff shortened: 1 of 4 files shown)


ContestUtil/src/org/icpc/tools/contest/util/cms/JsonToTSVConverter.java
```diff
@@ -184,7 +184,7 @@ public boolean accept(File f) {
 			groupList.sort((g1, g2) -> compare(g1.getId(), g2.getId()));
 			teamList.sort((t1, t2) -> compare(t1.getId(), t2.getId()));
 			orgList.sort((i1, i2) -> compare(i1.getId(), i2.getId()));
-			personList.sort((m1, m2) -> compare(m1.getTeamId(), m2.getTeamId()));
+			personList.sort((m1, m2) -> compare(m1.getId(), m2.getId()));
```
nickygerritsen, line 187: We could sort *first* by teamID and then by person ID if we still want teams together.

nickygerritsen (approved):

ContestUtil/src/org/icpc/tools/contest/util/cms/JsonToTSVConverter.java
deboer-tim: Thought about that, but wasn't sure it was worth the code :). Will see how this goes, can always switch to that.

PR #728, opened 2022-11-11
Decorated presentation window, to allow resizing

zalenski:


nickygerritsen (approved):

PresCore/src/org/icpc/tools/presentation/core/PresentationWindow.java
```diff
@@ -15,13 +14,14 @@
...
 import javax.imageio.ImageIO;
+import javax.swing.JFrame;
 
 import org.icpc.tools.presentation.core.internal.PresentationWindowImpl;
 
 /**
  * A window capable of displaying presentations.
  */
-public abstract class PresentationWindow extends Frame implements IPresentationHandler {
+public abstract class PresentationWindow extends JFrame implements IPresentationHandler {
```
zalenski, line 24: is this safe? any differences in fullscreen window behaviour (apart from properly maximizing on mac)?

> When I create a Swing JFrame instance instead, the same button toggles the window into fullscreen mode, which is the standard behaviour on MacOS.

https://stackoverflow.com/questions/70501295/change-awt-frames-maximize-behaviour-to-fullscreen-on-macos

PR #739, opened 2022-11-24
Move CDS config to PKCS12

deboer-tim:
Troy mentioned we should update the CDS default config to PKCS12, turns out the Liberty default keystore changed from JKS to PKCS12 a while ago.


deboer-tim:
FYI @troy2914

nickygerritsen (approved):

PR #746, opened 2023-02-16
Prefer SVG over any PNG when selecting images of a size.

nickygerritsen:


nickygerritsen:
We are not doing this for now, we need to think of what's best to do in this case.

PR #747, opened 2023-02-18
Fix some Java and style issues in presentation admin.

meisterT:
No functional changes intended.

### DIFF
--- a/PresAdmin/src/org/icpc/tools/presentation/admin/internal/DisplayAreaDialog.java
+++ b/PresAdmin/src/org/icpc/tools/presentation/admin/internal/DisplayAreaDialog.java
@@ -54,8 +54,6 @@ public class DisplayAreaDialog extends Dialog {
 		label.setText("Display area (x,y,w,h):");
 		label.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
 
-		label = new Label(comp, SWT.NONE);
-
 		final Text text = new Text(comp, SWT.BORDER);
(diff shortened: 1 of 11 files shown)


PresAdmin/src/org/icpc/tools/presentation/admin/internal/DisplayDialog.java
```diff
@@ -52,8 +52,6 @@ protected void createUI(Composite comp) {
 		label.setText("Display:");
 		label.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
 
-		label = new Label(comp, SWT.NONE);
```
deboer-tim, line 55: Ditto here.

PresAdmin/src/org/icpc/tools/presentation/admin/internal/DisplayAreaDialog.java
```diff
@@ -54,8 +54,6 @@ protected void createUI(Composite comp) {
 		label.setText("Display area (x,y,w,h):");
 		label.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
 
-		label = new Label(comp, SWT.NONE);
```
deboer-tim, line 57: Did you try running? This looks like it would mess up the columns.

deboer-tim (commented):
Most of these are code cleanup and language syntax that didn't exist when the code was written, all good except for the two places that look like they'll mis-align UI.

Issue #753, opened 2023-02-28
Missconfigured IP during contests


MarcoUCM:
In a contest without internet access the inferred IP that Presenter and other tools use depends on the order the network interfaces are listed (method `getLocalAddress` in org.icpc.tools.contest.model.internal.NetworkUtil (https://github.com/icpctools/icpctools/blob/main/ContestModel/src/org/icpc/tools/contest/model/internal/NetworkUtil.java).

Depending on the network infrastructure and the number of network interfaces the host has this may not be the desired IP.

I'd suggest to add a way to override this behavior. In a past contest, we managed to get it working extending the method: if an environment variable was present it used the IP stored on it.

See commit: https://github.com/icpctools/icpctools/compare/main...programame:icpctools:ip-from-envvar

If that sounds good we could PR it.

deboer-tim:
I really don't love this, as to me this is a case where something is broken and other things/tools would eventually fail too, and require additional workarounds... but that said, there is almost zero risk in adding this as a back door if it helps you.

I'd accept a PR, but could you rename the env var to something like "ICPC_TOOLS_IP"? (would make it consistent with at least one other var we have)

MarcoUCM:
Sure.
Thank you, I appreciate it :)

PR #754, opened 2023-03-01
Local IP for tools like Presenter may be read from env var

MarcoUCM:
See issue #753


deboer-tim (approved):
LGTM

deboer-tim:
Thanks for the contribution, merged!

PR #758, opened 2023-03-09
Adding Expected-to-Advance award

jifland:
From the command line this runs, builds the award with the teams associated with it, and then exits without adding the award to the data file.  I'm sure I'm missing something.

Also I would be interested in any feedback.


deboer-tim:
Sorry for the delay. I was finally able to take a look at this. The code is great, but I ended up making a couple minor changes and have proposed as PR #762. Assuming that gets integrated I'll close this.

deboer-tim:
PR #762 has been merged and includes these changes.

PR #768, opened 2023-04-06
Team label support

deboer-tim:
We're adding team labels to the Contest API, so we need to match this throughout the tools.

This PR covers:
- The actual API change.
- Default label to the id if it isn't present (this is a bit of a hack done when the name is set - couldn't think of a cleaner way). This ensures that clients can always rely on the label and it is there even if the CCS/contest package doesn't have one.
- Impact to all downstream clients: balloon printer, coach view, presentations, award util, and CDS web views.

Could I have missed something? Sure. I did a pretty thorough search which turned up all the impacts that I could think of, but until we test at NAC we won't know for sure. For now I switched all 'user output/messages' to labels, but left the few internal log messages as id because that's what you'll need for things like API lookup.

### DIFF
--- a/PresContest/src/org/icpc/tools/presentation/contest/internal/presentations/map/TeamIntroPresentation.java
+++ b/PresContest/src/org/icpc/tools/presentation/contest/internal/presentations/map/TeamIntroPresentation.java
@@ -190,7 +190,7 @@ public class TeamIntroPresentation extends AbstractICPCPresentation {
 				}
 				String label = t.getLabel() + " - " + t.getActualDisplayName();
 				if (showOrganizations) {
-					label = t.getLabel() + " - " + org.getFormalName();
+					label = org.getFormalName();
 				}
 				Position p = new Position(lon, lat, 1, label);
 				createOrgLogo(p, org, height);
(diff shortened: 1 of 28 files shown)


PresContest/src/org/icpc/tools/presentation/contest/internal/presentations/map/TeamIntroPresentation.java
```diff
@@ -187,9 +188,9 @@ public static GroupZoom setTargets(IContest contest, String groupId, int height,
 					minLon = Math.min(minLon, lon);
 					maxLon = Math.max(maxLon, lon);
 				}
-				String label = t.getId() + " - " + t.getActualDisplayName();
+				String label = t.getLabel() + " - " + t.getActualDisplayName();
 				if (showOrganizations) {
-					label = org.getFormalName();
+					label = t.getLabel() + " - " + org.getFormalName();
```
nickygerritsen, line 193: I think this is not what we want? If `showOrganizations` is set, we don't want team info do we?

nickygerritsen (approved):
One small comment.

PresContest/src/org/icpc/tools/presentation/contest/internal/presentations/map/TeamIntroPresentation.java
deboer-tim: I waffled when doing the change, but you're right. Removed this and merging.

PR #772, opened 2023-04-16
Better i8n support for resolver and es translation

MarcoUCM:
- The FTS award citation was hardcoded in the resolver in order to join the multiple FTS awards ("First to solve problema A, B and K"). Now it uses the messages.properties
- Added spanish translation


deboer-tim (approved):
LGTM

PR #813, opened 2023-07-28
Add support for showing (colored) contest title in most presentations.

### LINKED ISSUE #808: Resolver option to include contest title
The resolver should have a command line option to include the contest (short) name on all screens.

nickygerritsen:
Fixes #808

Some screenshots:
Screenshot 2023-07-28 at 11 47 17
Screenshot 2023-07-28 at 11 47 35
Screenshot 2023-07-28 at 11 47 54
Screenshot 2023-07-28 at 11 48 02
Screenshot 2023-07-28 at 11 48 20
Screenshot 2023-07-28 at 11 48 32

### DIFF
--- a/PresContest/src/org/icpc/tools/presentation/contest/internal/AbstractICPCPresentation.java
+++ b/PresContest/src/org/icpc/tools/presentation/contest/internal/AbstractICPCPresentation.java
@@ -1,7 +1,7 @@
 package org.icpc.tools.presentation.contest.internal;
 
-import java.awt.FontMetrics;
-import java.awt.Graphics2D;
+import java.awt.*;
+import java.awt.image.BufferedImage;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -13,12 +13,129 @@ import org.icpc.tools.presentation.core.Presentation;
 public abstract class AbstractICPCPresentation extends Presentation {
 	private IContest contest = ContestData.getContest();
 
+	private Font contestTitleFont;
+	private final int contestTitleHeight = 20;
+	private final int contestTitleMargin = 2;
+	private BufferedImage contestTitleImage;
+	private static String contestTitleTemplate;
+	protected static Color contestTitleColor;
+
+	protected void setupContestTitle() {
+		final float dpi = 96;
+		float size = (int) (height * 72.0 * 0.028 / dpi);
+		contestTitleFont = ICPCFont.deriveFont(Font.BOLD, size * 2.2f);
+
+		contestTitleImage = createContestTitleImage();
+	}
+
+	@Override
+	public void setSize(Dimension d) {
+		super.setSize(d);
+
+		contestTitleImage = null;
+
+		if (d.width == 0 || d.height == 0)
+			return;
+
+		setupContestTitle();
+	}
+
+	protected void paintContestTitle(Graphics2D g) {
+		g.drawImage(contestTitleImage, 0, 0, null);
+	}
+
+	protected BufferedImage createContestTitleImage() {
+		// The + 2 is here (and in paint(g)) to make sure there is some room between the header and the rest of the presentation
+		BufferedImage img = new BufferedImage(width, contestTitleHeight + 2 * contestTitleMargin + 2, Transparency.OPAQUE);
+		Graphics2D g = (Graphics2D) img.getGraphics();
+		g.setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
+		g.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
+		g.setRenderingHint(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY);
+
+		if (getContestTitle() != null) {
+			drawContestTitle(g);
+		}
+
+		g.dispose();
+		return img;
+	}
+
+	protected void drawContestTitle(Graphics2D g) {
+		String s = getContestTitle();
+
+		Color color = isLightMode() ? Color.WHITE : Color.BLACK;
+		if (contestTitleColor != null) {
+			color = contestTitleColor;
+		}
+		g.setColor(color);
+		g.fillRect(0, 0, width, contestTitleHeight + contestTitleMargin * 2);
+
+		g.setColor(ICPCColors.foregroundColor(color));
+		g.setFont(contestTitleFont);
+		FontMetrics fm = g.getFontMetrics();
+		g.drawString(s, (width - fm.stringWidth(s)) / 2, contestTitleMargin + fm.getAscent());
+	}
+
+	protected boolean shouldDrawContestTitle() {
+		return getContestTitle() != null;
+	}
+
+	public static void setContestTitleTemplate(String titleTemplate) {
+		AbstractICPCPresentation.contestTitleTemplate = titleTemplate;
+	}
+
+	public static void setContestTitleColor(Color titleColor) {
+		AbstractICPCPresentation.contestTitleColor = titleColor;
+	}
+
+	protected String getContestTitle() {
+		if (contestTitleTemplate == null) {
+			return null;
+		}
+
+		return contestTitleTemplate
+				.replace("{contest.name}", contest.getName())
+				.replace("{contest.formal_name}", contest.getActualFormalName());
+	}
+
+	@Override
+	public void paint(Graphics2D g) {
+		g.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
+		g.setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
+		g.setRenderingHint(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY);
+
+		Graphics2D g2 = g;
+
+		if (shouldDrawContestTitle()) {
+			paintContestTitle(g);
+
+			IContest c = getContest();
+			if (c == null || c.getNumTeams() == 0)
+				return;
+
+			g2 = (Graphics2D) g.create();
+			g2.translate(0, contestTitleMargin * 2 + contestTitleHeight + 2);
+			g2.setClip(0, 0, width, height - contestTitleMargin * 2 - contestTitleHeight - 2);
+		}
+
+		paintImpl(g2);
+
+		if (shouldDrawContestTitle()) {
+			g2.dispose();
+		}
+	}
+
+	protected abstract void paintImpl(Graphics2D g);
+
 	protected IContest getContest() {
 		return contest;
 	}
 
 	public void setContest(IContest newContest) {
 		contest = newContest;
+		if (width > 0 && height > 0) {
+			setupContestTitle();
+		}
 	}
 
 	/**
--- a/PresContest/src/org/icpc/tools/presentation/contest/internal/AbstractICPCPresentation.java
+++ b/PresContest/src/org/icpc/tools/presentation/contest/internal/AbstractICPCPresentation.java
@@ -14,8 +14,9 @@ public abstract class AbstractICPCPresentation extends Presentation {
 	private IContest contest = ContestData.getContest();
 
 	private Font contestTitleFont;
-	private final int contestTitleHeight = 20;
-	private final int contestTitleMargin = 2;
+	private int contestTitleHeight;
+	private final double contestTitleMarginPercentage = 0.1; // Percentage of contestTitleHeight
+	private int contestTitleMargin;
 	private BufferedImage contestTitleImage;
 	private static String contestTitleTemplate;
 	protected static Color contestTitleColor;
@@ -24,6 +25,10 @@ public abstract class AbstractICPCPresentation extends Presentation {
 		final float dpi = 96;
 		float size = (int) (height * 72.0 * 0.028 / dpi);
 		contestTitleFont = ICPCFont.deriveFont(Font.BOLD, size * 2.2f);
+
+		Canvas c = new Canvas();
+		contestTitleHeight = c.getFontMetrics(contestTitleFont).getHeight();
+		contestTitleMargin = (int)(contestTitleHeight * contestTitleMarginPercentage);
 
 		contestTitleImage = createContestTitleImage();
 	}
--- a/PresContest/src/org/icpc/tools/presentation/contest/internal/ClientLauncher.java
+++ b/PresContest/src/org/icpc/tools/presentation/contest/internal/ClientLauncher.java
@@ -1,5 +1,6 @@
 package org.icpc.tools.presentation.contest.internal;
 
+import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.util.List;
 
@@ -37,6 +38,13 @@ public class ClientLauncher {
 		System.out.println("     --multi-display p@wxh");
 		System.out.println("         Stretch the presentation across multiple clients. Use \"2@3x2\"");
 		System.out.println("         to indicate this client is position 2 (top middle) in a 3x2 grid");
+		System.out.println("     --contestTitle");
+		System.out.println("         Set the contest title of the presentations to the contest name");
+		System.out.println("     --contestTitleTemplate template");
+		System.out.println("         Set the contest title of the presentations using a template. Parameters:");
+		System.out.println("         {contest.name) and {contest.formal_name)");
+		System.out.println("     --contestTitleColor");
+		System.out.println("         Set the contest title color of the presentations. Currently supports `blue` and `yellow`");
 		System.out.println("     --light");
 		System.out.println("         Use light mode");
 		System.out.println("     --help");
@@ -52,6 +60,8 @@ public class ClientLauncher {
 		String[] nameStr = new String[1];
 		String[] displayStr = new String[2];
 		String[] displayName = new String[1];
+		String[] contestTitle = new String[1];
+		Color[] contestTitleColor = new Color[1];
 		boolean[] lightMode = new boolean[1];
 		ContestSource contestSource = ArgumentParser.parse(args, new OptionParser() {
 			@Override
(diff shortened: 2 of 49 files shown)


deboer-tim:
Some challenges we need to discuss:
- The charts don't extend AbstractICPCPresentation so we'd need to do those too.
- The height should be proportional instead of fixed, otherwise it'll be too big on small displays and too small on 4k.
- When the child presentations draw they have been translated down without them knowing about it. e.g. you can see half a scoreboard row missing on some of them, partial award text, and there are others like the WorldPresentation where the bottom text will likely be cutoff too. All of these would need to be adjusted to deal with the header.
- If we have non-contest-related presentations in a mix (e.g. some photos or logo), this header will only be on some screens. It wasn't clear to me if they expected it to 'always be there'.
- Shouldn't this be in the standalone launcher too?
- Help says 'Currently supports `blue` and `yellow`' but code looks like it does more?
- Unless it has just been decided and was buried somewhere in the slack chain, John couldn't give an answer on Wednesday about whether we'd ever have a single screen that would need to show the data from one contest, then the other contest. e.g. if there's a screen in the middle that we would cycle between the two. That would mean doubling the command-line args (--contestTitleColor1 --contestTitleColor2? or --contestTitleColor a,b?). I think we either need an answer on this, or need to plan for either.
- Likewise, the current plan for the resolver was to have a single resolver controlling running both. So does that mean doubling the args there as well?

nickygerritsen:
> Some challenges we need to discuss:
> 
> * The charts don't extend AbstractICPCPresentation so we'd need to do those too.

Any reason they are not extending AbstractICPCPresentation? I tried this and it seemed to still work.
> * The height should be proportional instead of fixed, otherwise it'll be too big on small displays and too small on 4k.

I copied this from `TitledPresentation` which then has the same issue? But it shouldn't be too hard to change the height dynamically? As long as we can also change the font size.
> * When the child presentations draw they have been translated down without them knowing about it. e.g. you can see half a scoreboard row missing on some of them, partial award text, and there are others like the WorldPresentation where the bottom text will likely be cutoff too. All of these would need to be adjusted to deal with the header.

I didn't check individual presentations yet. But since I am clipping and resizing, shouldn't that take care of most things? Or is the issue that some use `width`/`height` instead of the dimensions of the `Graphics2D`?
> * If we have non-contest-related presentations in a mix (e.g. some photos or logo), this header will only be on some screens. It wasn't clear to me if they expected it to 'always be there'.

Not clear to me either. We need to check with @clevengr probably. Same for the clock presentation where we explicitly hide it.
> * Shouldn't this be in the standalone launcher too?

Forgot that, thanks
> * Help says 'Currently supports `blue` and `yellow`' but code looks like it does more?

Whoops, I will fix this. Was legacy stuff
> * Unless it has just been decided and was buried somewhere in the slack chain, John couldn't give an answer on Wednesday about whether we'd ever have a single screen that would need to show the data from one contest, then the other contest. e.g. if there's a screen in the middle that we would cycle between the two. That would mean doubling the command-line args (--contestTitleColor1 --contestTitleColor2? or --contestTitleColor a,b?). I think we either need an answer on this, or need to plan for either.
> * Likewise, the current plan for the resolver was to have a single resolver controlling running both. So does that mean doubling the args there as well?

Right, I didn't handle these two cases yet since switching contests on the fly is something we need for both tools then. I'd say let's tackle that then. At least I made it such that the header image will be recalculated (and thus color changed) when you switch contests. We might want to also do this when you call the two setters for color/title template.

deboer-tim:
> > Some challenges we need to discuss:
> > 
> > * The charts don't extend AbstractICPCPresentation so we'd need to do those too.
> 
> Any reason they are not extending AbstractICPCPresentation? I tried this and it seemed to still work.

Historically there was a good reason. Do I remember what it was or if it still applies? No.

> > * The height should be proportional instead of fixed, otherwise it'll be too big on small displays and too small on 4k.
> 
> I copied this from `TitledPresentation` which then has the same issue? But it shouldn't be too hard to change the height dynamically? As long as we can also change the font size.
> 
> > * When the child presentations draw they have been translated down without them knowing about it. e.g. you can see half a scoreboard row missing on some of them, partial award text, and there are others like the WorldPresentation where the bottom text will likely be cutoff too. All of these would need to be adjusted to deal with the header.
> 
> I didn't check individual presentations yet. But since I am clipping and resizing, shouldn't that take care of most things? Or is the issue that some use `width`/`height` instead of the dimensions of the `Graphics2D`?

Graphics2D doesn't have a dimension, it's just a context. The presentations still think they have the full size available from setSize(). 

> > * If we have non-contest-related presentations in a mix (e.g. some photos or logo), this header will only be on some screens. It wasn't clear to me if they expected it to 'always be there'.
> 
> Not clear to me either. We need to check with @clevengr probably. Same for the clock presentation where we explicitly hide it.

Agreed.

> > * Shouldn't this be in the standalone launcher too?
> 
> Forgot that, thanks
> 
> > * Help says 'Currently supports `blue` and `yellow`' but code looks like it does more?
> 
> Whoops, I will fix this. Was legacy stuff
> 
> > * Unless it has just been decided and was buried somewhere in the slack chain, John couldn't give an answer on Wednesday about whether we'd ever have a single screen that would need to show the data from one contest, then the other contest. e.g. if there's a screen in the middle that we would cycle between the two. That would mean doubling the command-line args (--contestTitleColor1 --contestTitleColor2? or --contestTitleColor a,b?). I think we either need an answer on this, or need to plan for either.
> > * Likewise, the current plan for the resolver was to have a single resolver controlling running both. So does that mean doubling the args there as well?
> 
> Right, I didn't handle these two cases yet since switching contests on the fly is something we need for both tools then. I'd say let's tackle that then. At least I made it such that the header image will be recalculated (and thus color changed) when you switch contests. We might want to also do this when you call the two setters for color/title template.

FWIW I'm thinking contest color should maybe come from the Contest API so that it's set once and all tools can pick it up.

And yes, I am uneasy going down any road where I have no faith yet that this is actually what they want and won't change their mind next week. We will end up with a bunch of code changes (I mean this one is already massive) that are either used one time, or we need to back out and do something else. We need some confirmation that the critical bits like this are on the right track.

nickygerritsen:
> > I didn't check individual presentations yet. But since I am clipping and resizing, shouldn't that take care of most things? Or is the issue that some use `width`/`height` instead of the dimensions of the `Graphics2D`?
> 
> Graphics2D doesn't have a dimension, it's just a context. The presentations still think they have the full size available from setSize().

OK so probably we need some way to tell presentations their actual 'working area'? Perhaps we can set a `usableHeight` that subtracts the header height if set. But this would indeed mean finding all usages of `height` and adjusting. That's quite a bit of work so I'd prefer to only to this if we have confirmation this is what we want to use.

> > > * Unless it has just been decided and was buried somewhere in the slack chain, John couldn't give an answer on Wednesday about whether we'd ever have a single screen that would need to show the data from one contest, then the other contest. e.g. if there's a screen in the middle that we would cycle between the two. That would mean doubling the command-line args (--contestTitleColor1 --contestTitleColor2? or --contestTitleColor a,b?). I think we either need an answer on this, or need to plan for either.
> > > * Likewise, the current plan for the resolver was to have a single resolver controlling running both. So does that mean doubling the args there as well?
> > 
> > 
> > Right, I didn't handle these two cases yet since switching contests on the fly is something we need for both tools then. I'd say let's tackle that then. At least I made it such that the header image will be recalculated (and thus color changed) when you switch contests. We might want to also do this when you call the two setters for color/title template.
> 
> FWIW I'm thinking contest color should maybe come from the Contest API so that it's set once and all tools can pick it up.
> 
> And yes, I am uneasy going down any road where I have no faith yet that this is actually what they want and won't change their mind next week. We will end up with a bunch of code changes (I mean this one is already massive) that are either used one time, or we need to back out and do something else. We need some confirmation that the critical bits like this are on the right track.

If contest color comes from the contest API we can omit these settings indeed. We could extend the contest object with these internal properties easily right? If we also add `title_template` to it (if we need it and the name of the contest is not `46th ICPC World Finals` anyway), then we can drop the arguments.

zalenski:
> presentations their actual 'working area'

Hi! I had one initial thought, and then another one after looking more closely at the `Presentation`/`AbstractICPCPresentation` classes.

First thought was that you could layer the title as its own implementation, delegating to a sub-presentation that will be sized smaller. Similar to how you stack swing components. But I think that might require quite a bit of restructuring everywhere.

Second thought is that while there are hundreds of uses of `width` & `height` (they are plain protected fields, and used everywhere!), there are very few places where `setSize` is called to set them. An alternative to introducing a new `usableHeight`, could be instead to tweak the current `height` (and maybe introduce something like a `titleHeight`). The top level `setSize` implementation could simply split/adjust the total height given into title height and remaining height. Most setSize overrides I've seen so far call `super.setSize()` and then use `width/height` directly, so adjusting `height` like that should work.

deboer-tim:
> > presentations their actual 'working area'
> 
> Hi! I had one initial thought, and then another one after looking more closely at the `Presentation`/`AbstractICPCPresentation` classes.
> 
> First thought was that you could layer the title as its own implementation, delegating to a sub-presentation that will be sized smaller. Similar to how you stack swing components. But I think that might require quite a bit of restructuring everywhere.

FWIW, we used to have composite presentations that did exactly this. I don’t know if it’s still in the code vase but would be easy enough to reimplement - a parent presentation just paints the title and delegates to a child that was told about a smaller width or height.

Either way, I still want to confirm the requirements before we implement. 😉

clevengr:
> > > presentations their actual 'working area'
> > 
> > 
> > Hi! I had one initial thought, and then another one after looking more closely at the `Presentation`/`AbstractICPCPresentation` classes.
> > First thought was that you could layer the title as its own implementation, delegating to a sub-presentation that will be sized smaller. Similar to how you stack swing components. But I think that might require quite a bit of restructuring everywhere.
> 
> FWIW, we used to have composite presentations that did exactly this. I don’t know if it’s still in the code vase but would be easy enough to reimplement - a parent presentation just paints the title and delegates to a child that was told about a smaller width or height.
> 
> Either way, I still want to confirm the requirements before we implement. 😉

The best I can say about "the requirements" is the following:
- there needs to be a way to visually identify the contest to which a presentation applies.  The "colored title bar" approach is what is being used by ICPCLive, and so far I haven't heard anyone mention anything superior.
- if we're using colored title bars, there should be a way to use the same colors as those used by ICPCLive.  Currently I'm pretty certain Lidia is planning to use Blue and Yellow.
- colored title bars should not cut off content of the presentations.  (I noticed that in several of the screenshots above the little column titles at the top seem to be missing some pixels -- for example, `Rank`, `Name`, `Score` and `Time` in multiple screenshots seem to be cut off at the top.)
- Not directly related, but mentioned above:  I have confirmed that we Do need to be prepared to share screens between contests.  For example, WF46 teams will be on the left side of the room while WF47 teams will be on the right side, and there will be screens on the left side and screens on the right side.  However, the screens on the left side MIGHT be required to display WF47 presentations, and vice-versa.

Are there any other remaining questions about "requirements"?

nickygerritsen:
Yes: do we *always* show the colored title bars?
So also for for example the countdown presentation or let’s say the message presentation?

nickygerritsen:
> > [...]
> 
> The best I can say about "the requirements" is the following:
> 
> * there needs to be a way to visually identify the contest to which a presentation applies.  The "colored title bar" approach is what is being used by ICPCLive, and so far I haven't heard anyone mention anything superior.
> * if we're using colored title bars, there should be a way to use the same colors as those used by ICPCLive.  Currently I'm pretty certain Lidia is planning to use Blue and Yellow.

One discussion still left is how to render the bars, as discussed in Slack:

* A colored background, always white foreground. This is very hard to read but the point of @lperovskaya was that if we use the colors consistently, maybe being able to read the title is not the most important.
* A colored background with a text color that is white/black depending on what works best according to 'accessibility standardds'. In practice this makes the text black with the ICPC Yellow. @johnbrvc didn't really like this, called it 'old-fashioned nintendo inverse text'.
* A colored border instead of the whole background as suggested by @clevengr . I posted some samples in Slack and I am not a big fan but if other people like it, we can use it.

I don't care too much what we pick, but we should pick *something* that won't change anymore preferably.
> * colored title bars should not cut off content of the presentations.  (I noticed that in several of the screenshots above the little column titles at the top seem to be missing some pixels -- for example, `Rank`, `Name`, `Score` and `Time` in multiple screenshots seem to be cut off at the top.)

This is a bug with rendering on macOS in non-fullscreen mode. This is not normally there.
The issue that *is* there as described by Tim, 
> * Not directly related, but mentioned above:  I have confirmed that we Do need to be prepared to share screens between contests.  For example, WF46 teams will be on the left side of the room while WF47 teams will be on the right side, and there will be screens on the left side and screens on the right side.  However, the screens on the left side MIGHT be required to display WF47 presentations, and vice-versa.

So the remaining question is, do we *always* show the colored title bar? So also for example for:

* Countdown clocks (they should match for both contests).
* Messages ("Contest over, teams leave floor"), they might also match.
* Presentations that are not directly related to the current contest, like historical data.
> 
> Are there any other remaining questions about "requirements"?

I think with the above things answered we should be mostly there.

I do like the second idea of @zalenski . I think it's quite easy to implement and can implement it when we decided on above things.

clevengr:
> > > [...]
> > 
> > 
> > The best I can say about "the requirements" is the following:
> > 
> > * there needs to be a way to visually identify the contest to which a presentation applies.  The "colored title bar" approach is what is being used by ICPCLive, and so far I haven't heard anyone mention anything superior.
> > * if we're using colored title bars, there should be a way to use the same colors as those used by ICPCLive.  Currently I'm pretty certain Lidia is planning to use Blue and Yellow.
> 
> One discussion still left is how to render the bars, as discussed in Slack:
> 
> * A colored background, always white foreground. This is very hard to read but the point of @lperovskaya was that if we use the colors consistently, maybe being able to read the title is not the most important.

I'm not crazy about the idea of using something that is "very hard to read" and instead relying on the existence of the  color for the viewer to properly interpret it.  Viewers could be confused about what the colors represent (and what about color-blind viewers who need to see the text?)  I would suggest we scrap this particular variant.

> * A colored background with a text color that is white/black depending on what works best according to 'accessibility standardds'. In practice this makes the text black with the ICPC Yellow. @johnbrvc didn't really like this, called it 'old-fashioned nintendo inverse text'.

I agree with @johnbrvc; I don't care for the look.  However, I don't have a better suggestion...

> * A colored border instead of the whole background as suggested by @clevengr . I posted some samples in Slack and I am not a big fan but if other people like it, we can use it.

Having seen the samples, I'm also not a big fan; I retract the suggestion as not really viable.

> 
> I don't care too much what we pick, but we should pick _something_ that won't change anymore preferably.
> 
> > * colored title bars should not cut off content of the presentations.  (I noticed that in several of the screenshots above the little column titles at the top seem to be missing some pixels -- for example, `Rank`, `Name`, `Score` and `Time` in multiple screenshots seem to be cut off at the top.)
>
...
> ... The issue that _is_ there as described by Tim,
> 
> > * Not directly related, but mentioned above:  I have confirmed that we Do need to be prepared to share screens between contests.  For example, WF46 teams will be on the left side of the room while WF47 teams will be on the right side, and there will be screens on the left side and screens on the right side.  However, the screens on the left side MIGHT be required to display WF47 presentations, and vice-versa.

As I noted, we DO have to be prepared to deal with showing data from any contest on any screen.
> 
> So the remaining question is, do we _always_ show the colored title bar? So also for example for:
> 
> * Countdown clocks (they should match for both contests).
> * Messages ("Contest over, teams leave floor"), they might also match.
> * Presentations that are not directly related to the current contest, like historical data.
> 

I don't have a strong feeling that it's _necessary_ to _always_ show a colored title bar.  In some cases it does seem appropriate to NOT do so -- prime examples being Messages and "historical data".  In the case of Countdown clocks however, we DO need to be showing a colored title, just because we need to be prepared for the "Plan B" option that we get _forced_ to start the contests at different times (or make different adjustments to the clocks).  

But this raises another (somewhat minor, but real) question:  there is a presentation that shows a historical "Previous Year's First Solution"  (like "2017: MIT 6 min,  2018: ETH 5min,  2019 ITMO 7 min, ... 2023: ?? ")   If I recall correctly, this gets augmented on the fly when the first solution comes in.  If that's true, how are we going to deal with having TWO "First Solutions" (46 and 47)?

> > Are there any other remaining questions about "requirements"?
> 
> I think with the above things answered we should be mostly there.
> 
> I do like the second idea of @zalenski . I think it's quite easy to implement and can implement it when we decided on above things.

nickygerritsen:
> > 
> > * A colored background, always white foreground. This is very hard to read but the point of @lperovskaya was that if we use the colors consistently, maybe being able to read the title is not the most important.
> 
> I'm not crazy about the idea of using something that is "very hard to read" and instead relying on the existence of the color for the viewer to properly interpret it. Viewers could be confused about what the colors represent (and what about color-blind viewers who need to see the text?) I would suggest we scrap this particular variant.
> 
> > * A colored background with a text color that is white/black depending on what works best according to 'accessibility standardds'. In practice this makes the text black with the ICPC Yellow. @johnbrvc didn't really like this, called it 'old-fashioned nintendo inverse text'.
> 
> I agree with @johnbrvc; I don't care for the look. However, I don't have a better suggestion...
> 
> > * A colored border instead of the whole background as suggested by @clevengr . I posted some samples in Slack and I am not a big fan but if other people like it, we can use it.
> 
> Having seen the samples, I'm also not a big fan; I retract the suggestion as not really viable.

That basically only leaves the 'nintendo' option for now. If we agree this is most probably what we end up with, I can continue with this PR
> 
> > I don't care too much what we pick, but we should pick _something_ that won't change anymore preferably.
> > > * colored title bars should not cut off content of the presentations.  (I noticed that in several of the screenshots above the little column titles at the top seem to be missing some pixels -- for example, `Rank`, `Name`, `Score` and `Time` in multiple screenshots seem to be cut off at the top.)
> 
> ...
> 
> > ... The issue that _is_ there as described by Tim,
> > > * Not directly related, but mentioned above:  I have confirmed that we Do need to be prepared to share screens between contests.  For example, WF46 teams will be on the left side of the room while WF47 teams will be on the right side, and there will be screens on the left side and screens on the right side.  However, the screens on the left side MIGHT be required to display WF47 presentations, and vice-versa.
> 
> As I noted, we DO have to be prepared to deal with showing data from any contest on any screen.
> 
> > So the remaining question is, do we _always_ show the colored title bar? So also for example for:
> > 
> > * Countdown clocks (they should match for both contests).
> > * Messages ("Contest over, teams leave floor"), they might also match.
> > * Presentations that are not directly related to the current contest, like historical data.
> 
> I don't have a strong feeling that it's _necessary_ to _always_ show a colored title bar. In some cases it does seem appropriate to NOT do so -- prime examples being Messages and "historical data". In the case of Countdown clocks however, we DO need to be showing a colored title, just because we need to be prepared for the "Plan B" option that we get _forced_ to start the contests at different times (or make different adjustments to the clocks).

Which does mean that if clocks ARE the same, we still show a colored bar. Not a big issue I think, but just mentioning it.

Also I think it then makes sense to display the bar on as many presentations as we can easily implement. It doesn't hurt and is more consistent IMHO.
> 
> But this raises another (somewhat minor, but real) question: there is a presentation that shows a historical "Previous Year's First Solution" (like "2017: MIT 6 min, 2018: ETH 5min, 2019 ITMO 7 min, ... 2023: ?? ") If I recall correctly, this gets augmented on the fly when the first solution comes in. If that's true, how are we going to deal with having TWO "First Solutions" (46 and 47)?

Good catch. I think we do not currently do anything with this @deboer-tim ? So one screen shows only the FTS for 46 and one only shows the FTS for 47, right?
> 
> > > Are there any other remaining questions about "requirements"?
> > 
> > 
> > I think with the above things answered we should be mostly there.
> > I do like the second idea of @zalenski . I think it's quite easy to implement and can implement it when we decided on above things.

clevengr:
...
> That basically only leaves the 'nintendo' option for now. If we agree this is most probably what we end up with, I can continue with this PR

I don't have any better suggestions.  Perhaps @johnbrvc does?  If not, I suggest we default to the "nintendo" option as the "least bad choice"...

deboer-tim:
> > But this raises another (somewhat minor, but real) question: there is a presentation that shows a historical "Previous Year's First Solution" (like "2017: MIT 6 min, 2018: ETH 5min, 2019 ITMO 7 min, ... 2023: ?? ") If I recall correctly, this gets augmented on the fly when the first solution comes in. If that's true, how are we going to deal with having TWO "First Solutions" (46 and 47)?
> 
> Good catch. I think we do not currently do anything with this @deboer-tim ? So one screen shows only the FTS for 46 and one only shows the FTS for 47, right?

We set which FTSs for the presentation to use, not for the contest. So probably makes sense to use something like 43/44/45 as the history for both, even though 47 would have used 44/45/46 any other time.

johnbrvc:
> 

No, I can't think of anything else better.  I guess it's Nintendo mode.

nickygerritsen:
(in a Mario voice) Lets-a-go! 😅

deboer-tim:
Fixed another route via PR #880 and others.

PR #819, opened 2023-08-11
Use Contest for storing defaults

deboer-tim:
This change does two things:
- adds support for loading contest objects as default objects (required for a different PR that's coming).
- switches from storing default objects in an array to using a Contest.

'Defaults' are config loaded from disk that should always provide default properties from anything coming 'live'. This is what allows us to load things like org logos from disk even if the CCS doesn't support them. When we get the object from the CCS we simply add default values for any properties that weren't set.

Storing these objects in an array has always been a performance bottleneck, because if you loaded over 1000 items from disk (140 teams, 140 orgs, 800 people, 15 problems, 10 judgement types, some groups and languages), you need to scan through all 1000 items later to confirm that you didn't have any defaults to apply to an object. Using a Contest (with underlying ContestData storage) allows items to be looked up in a quick type-specific cache, far quicker. e.g. if you need to find a team you're only looking at the pool of 140 teams, or problems only perhaps 15.

I'm not going to lie; I'd rather make a change like this before a tiny contest that had lots of time to test... OTOH, it is obvious, was overdue, should be easy enough to catch potential regressions, and enables the other change I need to do. There's a <5% of regression, but also a strong likelihood of an order of magnitude improvement in time to load large contests.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/util/PlaybackContest.java
+++ b/CDS/src/org/icpc/tools/cds/util/PlaybackContest.java
@@ -59,7 +59,7 @@ public class PlaybackContest extends Contest {
 	protected boolean isTesting;
 	protected double timeMultiplier = Double.NaN;
 	protected Long startTime;
-	private List<IContestObject> defaults = new ArrayList<>();
+	private Contest defaultConfig = new Contest(false);
 	protected boolean configurationLoaded;
 
 	public PlaybackContest(ConfiguredContest cc) {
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

nickygerritsen:
> Storing these objects in an array has always been a performance bottleneck, because if you loaded over 1000 items from disk (140 teams, 140 orgs, 800 people, 15 problems, 10 judgement types, some groups and languages), you need to scan through all 1000 items later to confirm that you didn't have any defaults to apply to an object. Using a Contest (with underlying ContestData storage) allows items to be looked up in a quick type-specific cache, far quicker. e.g. if you need to find a team you're only looking at the pool of 140 teams, or problems only perhaps 15.

Isn't it even faster since the Contest object uses a map to find the correct object, so it's basically a constant time, right?

deboer-tim:
> > Storing these objects in an array has always been a performance bottleneck, because if you loaded over 1000 items from disk (140 teams, 140 orgs, 800 people, 15 problems, 10 judgement types, some groups and languages), you need to scan through all 1000 items later to confirm that you didn't have any defaults to apply to an object. Using a Contest (with underlying ContestData storage) allows items to be looked up in a quick type-specific cache, far quicker. e.g. if you need to find a team you're only looking at the pool of 140 teams, or problems only perhaps 15.
> 
> Isn't it even faster since the Contest object uses a map to find the correct object, so it's basically a constant time, right?

Yeah, that too. While writing it I was thinking 'I'm pretty sure it's even faster than this', but I couldn't remember why. 😂

PR #820, opened 2023-08-11
Color value for contests

deboer-tim:
Although it's ugly, IMHO the only way to implement #808 (contest banner in resolver and presentations) 'properly' is to set a color on the contest object, the same way that problems have colors (in fact, this implementation is copied from there).

In the contest configuration we add a file like `extend-color/contests.json` with contents like:
  `{"id":"sharm","rgb":"FF0000"}`
Now all clients see the correct color to use in the contest endpoint or feed. Presentations and resolver can use it for the contest-specific banner. No command line args required for color; clients that load two contests can see a value per-contest. Live can pick up the correct color too; or if they choose to set it differently, at least it is clear what we are using and when it doesn't match.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/IContest.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/IContest.java
@@ -1,5 +1,6 @@
 package org.icpc.tools.contest.model;
 
+import java.awt.Color;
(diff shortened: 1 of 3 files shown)


nickygerritsen (approved):
This is better than a command line argument for sure.

PR #822, opened 2023-09-02
Resolver info as contest data

deboer-tim:
Opening this as draft since this works standalone but still needs some minor changes to work through the CDS, and I haven't tested a couple things (e.g. predetermined json).

Resolver options/settings are currently passed from presenter through the CDS to clients as global presentation settings. The resolver presenter keeps track of what clients are connected to the CDS and broadcasts a message to them whenever they first connect or when settings (clicks, speed, etc) change. This is mostly historical; there was no other way to send data to all clients and a short presentation setting was quicker than non-standard chunks of XML contest data.

The Contest API now allows arbitrary data to be sent. This change creates a new singleton resolve-info contest object that contains all of the resolver logic settings (e.g. single step start row, predetermined steps), UI options (e.g. row offset), and current state (e.g. clicks, speed factor). When something changes the resolver presenter simply patches the contest object; all clients automatically receive it because it's part of the contest data.

Why make this change now? There are a few benefits:

- No more special-case/twisting how presentation control data should be used.
- Contest API now allows this; no more performance benefit of custom strings compared to ndjson as there was 15 years ago.
- Resolver options can optionally be preconfigured in contest data (or extend-data) now - i.e. stored in one spot instead of matching client and server options. They can be preconfigured in advance, and part of the contest archive.
- Clicks and scroll speed are now per-contest (vs global), so a single CDS can be used to resolve multiple independent contests.
- Predetermined steps are in regular json now.


nickygerritsen (approved):
Didn't check the code or logic in detail but if it works I believe you

PR #826, opened 2023-09-15
Images generation support for CMS CPF

deboer-tim:
The CMS is now exporting the Contest Package Format, which works great and means we now have a single download instead of 3, but means the format and location of images has completely changed.

A rough summary of the changes:
- Removed second CMS source folder. Now you just point this utility at the single contest location on disk. (woo hoo!)
- But this means that some logos are already 'in place'. So instead of always copying from the old CMS output folder to the Contest folder, we leave the file if it is good. If the name doesn't match the spec, we create a full size copy.
- Filter out resized images when looking for the original logo file.

Using this I was able to do a first pass at editing all the logos for the Sharm 46 & 47 contests. More work will likely need to be done, but even with some rough bits this is working better than previous years.

Support for team member photos is incomplete; I started making necessary changes but there is more work to do, similar to logos.

### DIFF
--- a/ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
+++ b/ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
@@ -54,6 +54,10 @@ public class ImagesGenerator {
 	private static final int HD_TOP = 30;
 	private static final double FUDGE = 0.075; // 7.5%
 
+	private static final String[] LOGO_EXTENSIONS = new String[] { "png", "svg", "jpg", "jpeg" };
+	// private static final String[] PHOTO_EXTENSIONS = new String[] { "jpg", "jpeg", "png", "svg"
+	// };
+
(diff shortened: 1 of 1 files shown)


ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
```diff
@@ -54,6 +54,10 @@ public class ImagesGenerator {
 	private static final int HD_TOP = 30;
 	private static final double FUDGE = 0.075; // 7.5%
 
+	private static final String[] LOGO_EXTENSIONS = new String[] { "png", "svg", "jpg", "jpeg" };
+	// private static final String[] PHOTO_EXTENSIONS = new String[] { "jpg", "jpeg", "png", "svg"
```
nickygerritsen, line 58: What's the commented line for?

nickygerritsen (approved):

Issue #812, opened 2023-07-26
HLS support in CDS


deboer-tim:
Both for streaming video outbound and incoming from teams.

nickygerritsen:
I was thinking / tinkering with this a bit.

First of all I found https://github.com/carlanton/m3u8-parser which can help us parse the playlists, which we probably need.

But our current VideoHandler setup does not really work well with HLS I think:

* Other video formats are one big stream, but this format has:
  * an `index.m3u8` (which should not really change)
  * one (or more `stream.m3u8`'s) that keep updating and contain links to the segmented TS/MP4 files
  * these TS/MP4 files, each one being a segment (the smallest possible thing HLS can stream)
* Other video have one mime type but HLS has 3: one for the m3u8's, one for the .ts and one for the .mp4.
* Recordings by default don't go in one file: we need a .m3u8 and one or more segments.

I tried tinkering a bit:

* We can add a `String getBaseStreamPostfix()` method to `VideoHandler` that we can use to determine the `href` in the API should be postfixed with that, for HLS that would be `index.m3u8`.
* We can add `List<String> allowedStreamPostfixes()` to say 'hey these things can be added after `/stream/0` for this format.

But that's about as far as I got. Some issues I see:

* `HLSHandler::validate` could check if the `index.m3u8` is valid but that's about it?
* `HLSHander::createReader` is already harder: we probably not only need an input stream but also the base URL so we can get the playlists. Also the `listener` doesn't really work anymore with just writing bytes etc., since it depends on what the client/recorder wants.

@deboer-tim do you have any thoughts how we could better make this more generic?

deboer-tim:
Last night I was able to stream HLS through the CDS for the first time (woo hoo!) with a _lot_ of limitations, e.g.several hacks and hardcoding in order to inject things at the right place, and m3u8 is essentially still a pass-through request.

I'm hoping to clean things up enough to release the first PR in a few days. After this there is still lots to do:
- Video file streaming is currently 'blocking' - the CDS caches all referenced video files before it returns the m3u8. This causes a delay on initial connection. It should be possible to return the m3u8 earlier and live stream as each file is being pulled. (or do we wait until each file is being pulled to go grab it? would take less memory and potentially less impact on team machine, but would impact latency)
- Video filenames or URLs and caching need to be unique for each stream.
- Test if current codec (on team machines) supports merging files to create reaction videos. Code it...
- Test if coach view works.
- m3u8 files need to be cached.
[...]

PR #832, opened 2023-10-03
Initial HLS support

deboer-tim:
This is the initial batch of changes to support HLS. It makes the structural/URL-facing changes required to support file-serving video formats (vs streaming), and sets the default video format to HLS (even though it's not really working yet!).

There is a *lot* of change in order to support HLS so I need to start pushing a branch, but there are numerous limitations (and likely some I still don't know about) listed in issue #812 and at the top of the HLSHandler. The current code works for one stream at a time, and hasn't been tested except through Safari.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/video/containers/HLSFileCache.java
+++ b/CDS/src/org/icpc/tools/cds/video/containers/HLSFileCache.java
@@ -1,0 +1,161 @@
+package org.icpc.tools.cds.video.containers;
+
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLConnection;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.icpc.tools.contest.Trace;
+import org.icpc.tools.contest.model.feed.HTTPSSecurity;
+
+public class HLSFileCache {
+	static class CachedFile {
+		public String name;
+		public byte[] b;
+		public long lastAccess;
+	}
+
+	private Map<String, CachedFile> map = new HashMap<>();
+	private List<String> preload = new ArrayList<>();
+
+	public HLSFileCache() {
+		// create
+	}
+
+	public void cleanCache() {
+		List<String> remove = new ArrayList<>(100);
+
+		long now = System.currentTimeMillis();
+		synchronized (map) {
+			for (String s : map.keySet()) {
+				CachedFile cf = map.get(s);
+				// older than 20s
+				if (cf.lastAccess < now - 20 * 1000L)
+					remove.add(s);
+			}
+		}
+
+		for (String s : remove) {
+			map.remove(s);
+		}
+	}
+
+	public boolean contains(String name) {
+		return map.containsKey(name);
+	}
+
+	public void preload(String name) {
+		if (map.containsKey(name) || preload.contains(name))
+			return;
+		preload.add(name);
+	}
+
+	public boolean hasPreload(String name) {
+		return preload.contains(name);
+	}
+
+	public void cache(String name, InputStream in) throws IOException {
+		ByteArrayOutputStream bout = new ByteArrayOutputStream();
+		BufferedInputStream bin = new BufferedInputStream(in);
+		byte[] b = new byte[1024 * 8];
+		int n = bin.read(b);
+		while (n != -1) {
+			bout.write(b, 0, n);
+			n = bin.read(b);
+		}
+
+		bin.close();
+		bout.close();
+
+		CachedFile cf = new CachedFile();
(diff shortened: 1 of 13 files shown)


CDS/src/org/icpc/tools/cds/video/containers/HLSFileCache.java
```diff
@@ -0,0 +1,161 @@
...
+
+		cf.lastAccess = System.currentTimeMillis();
+
+		BufferedOutputStream bout = new BufferedOutputStream(out);
+		bout.write(cf.b, 0, cf.b.length);
+		bout.close();
+	}
+
+	protected void cacheIt(String url, String name) {
+		// TODO: what are these gap files and do they matter?
```
nickygerritsen, line 99: From what I have done with HLS I haven't seen gap.mp4's, but my guess is that this happens when the source stream (i.e. to the webcam) is broken for a bit. Does the m3u8 have an `ext-x-discontinuity` above it?

CDS/src/org/icpc/tools/cds/video/containers/HLSFileCache.java
```diff
@@ -0,0 +1,161 @@
...
+
+		// already cached, don't grab them again
+		if (contains(name))
+			return;
+
+		try {
+			URLConnection conn = HTTPSSecurity.createURLConnection(new URL(url + "/" + name), null, null);
+			conn.setConnectTimeout(15000);
+			conn.setReadTimeout(10000);
+			conn.setRequestProperty("Content-Type", "video/mp4");
```
nickygerritsen, line 111: Do we need to set the content-type request header when requesting a video? Normally content-type is a return header, right?

CDS/src/org/icpc/tools/cds/video/containers/HLSFileCache.java
```diff
@@ -0,0 +1,161 @@
...
+
+			cache(name, conn.getInputStream());
+			System.out.println("  Cache success: " + name);
+		} catch (Exception e) {
+			System.err.println("  Cache fail: " + name);
+			Trace.trace(Trace.ERROR, "Could not cache HLS " + url + "/" + name, e);
+		}
+	}
+
+	protected void preloadIt(String url, String name) {
```
nickygerritsen, line 130: Isn't this a full duplicate of the method above?

CDS/src/org/icpc/tools/cds/video/containers/HLSHandler.java
```diff
@@ -0,0 +1,270 @@
...
+import org.icpc.tools.cds.video.VideoStreamHandler.IStreamListener;
+import org.icpc.tools.contest.Trace;
+import org.icpc.tools.contest.model.feed.HTTPSSecurity;
+
+/**
+ * HLS handler. https://localhost:8443/stream/11?_HLS_msn=10&_HLS_part=3&_HLS_skip=YES
+ * https://localhost:8443/stream/11?_HLS_msn=10&_HLS_part=4&_HLS_skip=YES
+ */
+// TODO - cache HLS m3u8
+// TODO - handle HLS query params
```
nickygerritsen, line 31: These are new for me, seems they are introduced when they introduced low latency HLS: https://developer.apple.com/documentation/http-live-streaming/enabling-low-latency-http-live-streaming-hls#Add-Low-Latency-HLS-Delivery-Directives

nickygerritsen (approved):
Read through the code and added some small comments. I didn't actually test the code since it will have many more changes anyway

CDS/src/org/icpc/tools/cds/video/containers/HLSFileCache.java
deboer-tim: It's typically near the top, looks like this:
```
#EXTM3U
#EXT-X-VERSION:9
#EXT-X-TARGETDURATION:1
#EXT-X-SERVER-CONTROL:CAN-BLOCK-RELOAD=YES,PART-HOLD-BACK=0.50000,CAN-SKIP-UNTIL=6.00000
#EXT-X-PART-INF:PART-TARGET=0.20000
#EXT-X-MEDIA-SEQUENCE:4
#EXT-X-SKIP:SKIPPED-SEGMENTS=2
#EXT-X-GAP
#EXTINF:1.00000,
gap.mp4
...
```
haven't found much spec-wise yet, but sounds like it's a marker that shouldn't be loaded nor ever requested by the client. I should likely switch to looking for the #EXT-X-GAP vs filtering by name.

CDS/src/org/icpc/tools/cds/video/containers/HLSFileCache.java
deboer-tim: Yes. Need to decide how to handle files that we expect to be ready vs those that we expect to "block until available unless they aren't" (preloads), so will likely need different code paths but haven't done it yet.

CDS/src/org/icpc/tools/cds/video/containers/HLSHandler.java
deboer-tim: Yes. Looks like It essentially gives you a way to block until content is available, so CDS will need to be intelligent about this too.

PR #834, opened 2023-10-06
Add submission files and entry point for analysts.

### LINKED ISSUE #831: Analyst role does not get a link to the submission during the contest
System setup:
CDS (prerelease version) reads from CMS (DOMjudge)

The mentioned event is before the freeze which the analyst role should be able to see.
Submission event as CDS admin (important, the href is there):
`{"type":"submissions","id":"587","data":{"id":"587","problem_id":"stroopwafel","team_id":"30092305","language_id":"cpp","files":[{"href":"contests/contest2023/submissions/587/files","filename":"files.zip","mime":"application/zip"}],"contest_time":"0:07:12.234","time":"2023-09-30T13:07:12.234+02:00"},"token":"cdD103"}`

Eventfeed entry voor CDS analyst user:
`{"type":"submissions","id":"587","data":{"id":"587","problem_id":"stroopwafel","team_id":"30092305","language_id":"cpp","contest_time":"0:07:12.234","time":"2023-09-30T13:07:12.234+02:00"},"token":"cd875"}
{"type":"judgements","id":"754","data":{"id":"754","submission_id":"587","start_contest_time":"0:07:12.875","start_time":"2023-09-30T13:07:12.875+02:00"},"token":"cd876"}`

nickygerritsen:
Fixes #831.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/internal/account/AnalystContest.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/internal/account/AnalystContest.java
@@ -72,8 +72,11 @@ public class AnalystContest extends SpectatorContest {
 	protected ISubmission filterSubmission(ISubmission sub) {
 		Submission s = (Submission) ((Submission) sub).clone();
 
-		if (!isBeforeFreeze(s))
+		if (!isBeforeFreeze(s)) {
+			s.setFiles(null);
+			s.add("entry_point", null);
 			s.setReaction(null);
+		}
 		return s;
 	}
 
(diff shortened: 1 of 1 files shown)


deboer-tim (changes requested):
This goes too far the other way and gives access to the file (just the reference, but I'd prefer not tell clients the URL they're not allowed to get) and entry point after the freeze, which is not allowed according to the policy. Can you add `s.add("entry_point", null)` and `s.setFiles(null)` in that if statement too?

nickygerritsen:
> This goes too far the other way and gives access to the file (just the reference, but I'd prefer not tell clients the URL they're not allowed to get) and entry point after the freeze, which is not allowed according to the policy. Can you add `s.add("entry_point", null)` and `s.setFiles(null)` in that if statement too?

Ah totally my bad, I thought analysts didn't get post-freeze submissions but that is dumb

nickygerritsen:
Done!

PR #835, opened 2023-10-06
HLS memory cache cleanup & X-GAP support

deboer-tim:
Next pass at HLS support:
- Changed the VideoAggregator so it doesn't create a massive thread pool if we have a serving handler.
- Changed VideoStream to pass execution requests back to the aggregator instead of holding onto a reference to the thread pool.
- Implemented a background thread to clean up the HLS cache, currently running every 5s and removing any file that hasn't been downloaded (in either direction) for 20s. [*]
- Added support for HLS X-GAP, so that we're not dependent on the name of the gap file.

[*] On my machine the cache for a single stream can get up to 27Mb, which if you do the math would run out of CDS memory in Sharm. But, we have no idea what the bitrate is that we're going to use or how many concurrent streams, so we can tune the cache or deal with this later...

### DIFF
--- a/CDS/src/org/icpc/tools/cds/video/containers/HLSHandler.java
+++ b/CDS/src/org/icpc/tools/cds/video/containers/HLSHandler.java
@@ -12,11 +12,13 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.icpc.tools.cds.service.ExecutorListener;
+import org.icpc.tools.cds.video.VideoAggregator;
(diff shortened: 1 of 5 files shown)


CDS/src/org/icpc/tools/cds/video/containers/HLSHandler.java
```diff
@@ -29,16 +31,20 @@
  */
 // TODO - cache HLS m3u8
 // TODO - handle HLS query params
-// TODO - no hardcoded stream
-// TODO - multiplexing
-// TODO - cache cleanup
+// TODO - multiplexing m3u8
+// TODO - switch gap.mp4 to using #EXT-X-GAP
```
nickygerritsen, line 35: This is not a todo anymore is it?

nickygerritsen (approved):

CDS/src/org/icpc/tools/cds/video/containers/HLSHandler.java
deboer-tim: Yeah, I missed that. I have the next PR almost ready, I'll fix it there.

PR #850, opened 2023-10-23
Resolver clients shouldn't have control

deboer-tim:
Due to the refactoring of the resolver code, client is loaded later and would sometimes be null at line 629, which meant non-presenter clients would have keyboard control. It wouldn't send events in, but it could get things out of sync until the presenter does something. This change makes it clear that isPresenter is the only indicator of control.

Fixed one ERROR to INFO.


deboer-tim:
@nickygerritsen could you still review?

nickygerritsen (approved):

PR #861, opened 2024-03-17
Send back empty JSON object for singletons.

### LINKED ISSUE #860: Contest state should return {}, not empty response, before any state changes.
Currently the contest state endpoint returns an empty response when the contest hasn't started yet. I think it should return `{}`.

nickygerritsen:
Fixes #860.

This make it that when we don't have a singleton object, we return `{}`.
Does this make sense for all singletons?

### DIFF
--- a/CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
+++ b/CDS/src/org/icpc/tools/cds/service/ContestRESTService.java
@@ -371,6 +371,9 @@ public class ContestRESTService extends HttpServlet {
 				writer.write(obj);
 			}
 		}
+		if (IContestObject.isSingleton(type) && objects.length == 0) {
+			writer.writeEmpty();
+		}
 		if (isArray)
 			writer.writePostlude();
 	}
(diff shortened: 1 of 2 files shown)


deboer-tim (changes requested):
I think this works for state, but is odd for things like the extended contest support, e.g. resolver and floor map - e.g. if I haven't configured a map it should probably be null; if I don't have access to the current resolving process running live then I probably shouldn't know it exists.

nickygerritsen:
> I think this works for state, but is odd for things like the extended contest support, e.g. resolver and floor map - e.g. if I haven't configured a map it should probably be null; if I don't have access to the current resolving process running live then I probably shouldn't know it exists.

So are you suggesting to hardcode the state check here?

deboer-tim:
> So are you suggesting to hardcode the state check here?

I think state probably needs a default/initial value in a contest.

nickygerritsen:
> > So are you suggesting to hardcode the state check here?
> 
> I think state probably needs a default/initial value in a contest.

Yeah that was my other idea. Feel free to implement that or I will try this at some point. Not sure (yet) where we would do this to get it working.

nickygerritsen:
Replaced by #875

PR #873, opened 2024-03-29
Generate multiple icons, so we can have 56x56 and 64x64.

nickygerritsen:
DOMjudge prefers 64x64.

### DIFF
--- a/ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
+++ b/ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
@@ -60,6 +60,7 @@ public class ImagesGenerator {
 
 	private static final Dimension DESKTOP = new Dimension(1920, 1080);
 	private static final Dimension ICON = new Dimension(56, 56);
+	private static final Dimension DOMJUDGE_ICON = new Dimension(64, 64);
 	private static final Dimension TILE = new Dimension(160, 160);
 
 	private static BufferedImage icpcLogo, logo;
@@ -435,7 +436,7 @@ public class ImagesGenerator {
 
 	public void generateOrganizationLogos() {
 		ImageSpec[] spec = new ImageSpec[] { new ImageSpec(DESKTOP, false), new ImageSpec(ICON, true),
-				new ImageSpec(TILE, true) };
+				new ImageSpec(DOMJUDGE_ICON, true), new ImageSpec(TILE, true) };
 		generateImages("organizations", "logo", "png", spec);
 	}
 


ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
```diff
@@ -434,8 +437,11 @@ private static File getDefaultFile(File folder, String property) {
 	}
 
 	public void generateOrganizationLogos() {
-		ImageSpec[] spec = new ImageSpec[] { new ImageSpec(DESKTOP, false), new ImageSpec(ICON, true),
-				new ImageSpec(TILE, true) };
+		ImageSpec[] spec = new ImageSpec[] { new ImageSpec(DESKTOP, false), new ImageSpec(TILE, true) };
+		spec = Arrays.copyOf(spec, spec.length + ICONS.length);
```
meisterT, line 441: nit: if you record the original length in a variable before the array copy, the indexing below will be nicer

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
```diff
@@ -434,8 +437,11 @@ private static File getDefaultFile(File folder, String property) {
...
+		for (int i = 0; i < ICONS.length; i++) {
+			spec[spec.length - ICONS.length + i] = new ImageSpec(ICONS[i], true);
+		}
```
nickygerritsen, line 444: If we can do this more cleanly in Java, I'm all up for it. Copilot suggested this ;)

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
nickygerritsen: Smart, done!

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
```diff
@@ -59,7 +59,10 @@ public class ImagesGenerator {
 	private static final Rectangle TEXT = new Rectangle(370, 795, 1275, 230);
 
 	private static final Dimension DESKTOP = new Dimension(1920, 1080);
-	private static final Dimension ICON = new Dimension(56, 56);
+	private static final Dimension[] ICONS = new Dimension[]{
```
deboer-tim, line 62: Why make icons an array instead of just adding a "DOMJUDGE_ICON" (naming tbd)?

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
```diff
@@ -434,8 +437,12 @@ private static File getDefaultFile(File folder, String property) {
 	}
 
 	public void generateOrganizationLogos() {
-		ImageSpec[] spec = new ImageSpec[] { new ImageSpec(DESKTOP, false), new ImageSpec(ICON, true),
-				new ImageSpec(TILE, true) };
+		ImageSpec[] spec = new ImageSpec[] { new ImageSpec(DESKTOP, false), new ImageSpec(TILE, true) };
+		int originalLength = spec.length;
+		spec = Arrays.copyOf(spec, spec.length + ICONS.length);
+		for (int i = 0; i < ICONS.length; i++) {
+			spec[originalLength + i] = new ImageSpec(ICONS[i], true);
```
deboer-tim, line 444: This will give you 64x64 padded icons and normally-scaled icons (maybe 64x50 or 57x64), but I thought you said you didn't want that. Or that's fine, you can deal with both/will ignore one?

deboer-tim (commented):
Just curious - I thought the 56x56 'spec required' sizes were ok in previous years, is there a reason/layout why they don't now? Should we be changing/making a spec change in future?

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
meisterT: In DOMjudge we prefer the 64x64 padded versions and ignore the rest if it exists

meisterT:
> Just curious - I thought the 56x56 'spec required' sizes were ok in previous years, is there a reason/layout why they don't now? Should we be changing/making a spec change in future?

In DOMjudge we have used 64x64 for a couple years now, but we did generate them ourselves.

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
nickygerritsen: Because I couldn't come up with a clean name besides exactly DOMJUDGE_ICON. If you are fine with that, I can do it.

nickygerritsen:
> > Just curious - I thought the 56x56 'spec required' sizes were ok in previous years, is there a reason/layout why they don't now? Should we be changing/making a spec change in future?
> 
> In DOMjudge we have used 64x64 for a couple years now, but we did generate them ourselves.

Yeah until last year we manually did this but now we can import them.

Do we know if anyone needs the 56x56? Otherwise we could also replace them.

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
nickygerritsen: There is no option to NOT output the normally-scaled ones right? We indeed need the padded ones.

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
deboer-tim: Feel free to using something like BIGGER_ICON if you prefer, but at least that name doesn't need a comment to remember why it's there. 😆

FWIW, I had this change ready 30s after we chatted about it, but didn't PR b/c I thought you said you didn't want the non-padded ones and hadn't had time to investigate.

deboer-tim:
> Do we know if anyone needs the 56x56? Otherwise we could also replace them.

They are spec required - hence my question asking if you think the spec should really be changed instead of having both.

nickygerritsen:
> > Do we know if anyone needs the 56x56? Otherwise we could also replace them.
> 
> They are spec required - hence my question asking if you think the spec should really be changed instead of having both.

If it's in the spec let's postpone that discussion for after Luxor.

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
nickygerritsen: Yeah we don't need the non-padded ones but they don't hurt I guess. We could change the `pad` boolean to an enum (later) so we can skip them.

nickygerritsen:
Added a hardcoded DOMjudge icon now.

deboer-tim (approved):

PR #880, opened 2024-04-10
Presentation contest branding (banner)

### LINKED ISSUE #808: Resolver option to include contest title

deboer-tim:
Allows a 'branding presentation' to be applied to presentation clients via an ICPC_BRANDING_PRESENTATION in a system property or environment variable. This presentation could do anything, but the supplied default one adds a banner across the top of the screen with the contest name and color.

For now this is per-client via the system property/env var to see how it works. If it does, we can expose through a command line property or admin control, or make different branding presentations.

Fixes #808.
Replaces #813.


nickygerritsen (approved):
This logic locks it to a branding presentation but that’s fine for now I guess

PR #886, opened 2024-04-13
Resolver scrolling speed control

deboer-tim:
Adds [ { and } ] keys to control the speed of scrolling in the resolver, for both the main scoreboard and team award list.

In the future we may want to:
 - combine this with +/-, i.e. you use the same keys but it only affects the screen you're on.
 - separate the main scoreboard scrolling from team list award scrolling.

but this adds the basic support that allows it to work, which we can expand on later.

Fixes #108.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/IResolveInfo.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/IResolveInfo.java
@@ -33,7 +33,7 @@ public interface IResolveInfo extends IContestObject {
 	double getSpeedFactor();
 
 	/**
-	 * Returns the scroll speed factor. Normal speed is 1, <1 is faster and >1 is slower.
+	 * Returns the scroll speed factor. Normal speed is 1, >1 is faster and <1 is slower.
 	 *
 	 * @return the relative scrolling speed
 	 */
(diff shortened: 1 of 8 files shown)


ContestModel/src/org/icpc/tools/contest/model/IResolveInfo.java
```diff
@@ -32,6 +32,13 @@ public String toString() {
 	 */
 	double getSpeedFactor();
 
+	/**
+	 * Returns the scroll speed factor. Normal speed is 1, <1 is faster and >1 is slower.
```
zalenski, line 36: In `TeamIntroPresentation`, `timeFactor` > 1 is faster and <1 is slower. There, timeFactor 2 is double the speed. Here, speedFactor .5 is half the time.

zalenski: it's actually the opposite here as well :) >1 is faster, <1 is slower

zalenski, line 36:
```suggestion
	 * Returns the scroll speed factor. Normal speed is 1, >1 is faster and <1 is slower.
```

zalenski (approved):

ContestModel/src/org/icpc/tools/contest/model/IResolveInfo.java
deboer-tim: Fixed.

deboer-tim: Fixed.

PR #895, opened 2024-04-14
Add force dark mode menu option to Admin

zalenski:


PresAdmin/src/org/icpc/tools/presentation/admin/internal/ClientsControl.java
```diff
@@ -83,6 +84,7 @@ public class ClientsControl extends Canvas {
 	protected int yOrigin = 0;
 	private boolean waitingForRedraw;
 	protected List<String> filter = new ArrayList<>();
+	protected boolean forceDark;
```
deboer-tim, line 87: Could you make this something like a `Boolean forceColorMode`? We don't need to support overriding to light mode now, but at least it would be generic and allow for it later.

PresAdmin/src/org/icpc/tools/presentation/admin/internal/ClientsControl.java
```diff
@@ -835,6 +845,10 @@ protected void setSmallThumbnails(boolean smallThumbs) {
 		resize();
 	}
 
+	protected void setForceDark(boolean forceDark) {
```
deboer-tim, line 848: Same here.

PresAdmin/src/org/icpc/tools/presentation/admin/internal/ClientsControl.java
zalenski: added light mode too

deboer-tim (approved):

PresAdmin/src/org/icpc/tools/presentation/admin/internal/ClientsControl.java
deboer-tim: thanks!

PR #899, opened 2024-04-15
Default contest id from yaml

deboer-tim:
If a contest is configured in a folder where the contest id doesn't match the folder name, and there is no CCS configured, and the contest id isn't set in cdsConfig.xml, we should be picking up the default contest id from the contest.yaml, not the folder name.

Adding support for this only for contest.yaml, not contest.json for now. We could wait until the contest is loaded and let the disk contest source load the id, but this delays too much, e.g. clients might be able to see the endpoints or contest exist in the web pages first.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
+++ b/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
@@ -303,8 +303,8 @@ public class ConfiguredContest {
 			}
 		}
 
-		// if no id, default to id from contest.yaml
-		// we could wait for the contest to load, but
+		// if no id, default to id from contest.yaml. We could wait for the contest to load,
+		// but by then some data could be accessible to clients
 		if (id == null && path != null) {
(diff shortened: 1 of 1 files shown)


CDS/src/org/icpc/tools/cds/ConfiguredContest.java
```diff
@@ -301,6 +302,18 @@ protected ConfiguredContest(Element e) {
 					id = url.substring(ind + 1);
 			}
 		}
+
+		// if no id, default to id from contest.yaml
+		// we could wait for the contest to load, but
```
nickygerritsen, line 307: but what?

nickygerritsen (approved):

CDS/src/org/icpc/tools/cds/ConfiguredContest.java
deboer-tim: Fixed.

PR #952, opened 2024-04-22
Presentation scoreboard support for scoring contests

deboer-tim:
Support for scoring contests by making the individual scoreboard cells show the score for the problem, and changing the Solved column to Score.


deboer-tim:
Realized I also need to remove penalty time, closing for now.

PR #960, opened 2024-05-22
Make pass/fail the default scoring algorithm.

nickygerritsen:
Without this the resolver assumed scoring as the default, meaning nothing showed up.


troy2914 (approved):
Looks good to me

PR #963, opened 2024-05-23
For scoring scoreboards always use the score, never the number of submissions

nickygerritsen:


deboer-tim (approved):

niemela:
This was tested at ICPC NAC NSA Challenge 2024, and worked great!

nickygerritsen:
I pushed a final commit that checks if a score is already the max score for a problem and then doesn't consider any submissions anymore. With that I think we can merge this. I will do that in a few days unless somebody opposes.

deboer-tim (approved):

PR #972, opened 2024-05-25
Do not generate awards with empty team lists

### LINKED ISSUE #971: Do not generate awards with empty team lists
Currently we generate awards with empty team lists, for example for FTS and for group winners. We should remove those.

nickygerritsen:
Fixes #971


nickygerritsen:
Marking this as draft since I think the discussion in #971 makes sense.

PR #973, opened 2024-05-25
Remove options that would start the contest too soon.

meisterT:
Everything less than 30s is outside of spec, and 30s is questionable.

### DIFF
--- a/CDS/WebContent/WEB-INF/jsps/countdown-control.jsp
+++ b/CDS/WebContent/WEB-INF/jsps/countdown-control.jsp
@@ -48,10 +48,6 @@
             <tr>
                 <td>
                     <select id="timeSelect" class="custom-select">
-                        <option value="0:00:01">1 second</option>
-                        <option value="0:00:05">5 seconds</option>
-                        <option value="0:00:15">15 seconds</option>
-                        <option value="0:00:30">30 seconds</option>
                         <option value="0:01:00" selected>1 minute</option>
                         <option value="0:05:00">5 minutes</option>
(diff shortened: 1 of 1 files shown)


meisterT:
[Spec:](https://ccs-specs.icpc.io/draft/contest_api#modifying-contests)

> The request should fail with a 403 error code if the contest is started or within 30s of starting, or if the new start time is in the past or within 30s.

deboer-tim (commented):
This would make it really hard to (eg) add or remove 10s from the countdown. Blocking below 30s should be done on the CDS or CCS.

meisterT:
Do you ever use add/remove with that little time?

deboer-tim:
Well we _have_ done that, e.g. I've been at a couple contests where the countdown was paused at like 2m if there were unresolved issues and then resumed again, and if the presenter didn't stop at exactly the 2m mark they added/removed a second to fix it instead of 'setting' it to 2m.

Ultimately it's the contest director/pausers choice of how they want to control things, so if John says he'd never do this and you know of enough other contests I'm fine removing these options, just want to be clear that it applies to adding/removing time too so it is unrelated to the spec 30s.

niemela:
> Everything less than 30s is outside of spec, and 30s is questionable.

Changing *to* a time less than 30s is not allowed, but changing *by* any amount (as long as it does not do the former) is.

Unless I'm misunderstanding the code, this change is a little bit too aggressive.

nickygerritsen:
Replaced by #1023

PR #985, opened 2024-05-27
Pass speed to inner resolvers

### LINKED ISSUE #982: Fix `--speed` for the resolver
The `--speed` option currently does nothing for the resolver, at least when combined with `--presenter` and not running against a CDS.

It is possible that it would work without `--presenter`, and this might be solved after #959 is fixed (?).

nickygerritsen:
Fixes #982

### DIFF
--- a/Resolver/src/org/icpc/tools/resolver/Resolver.java
+++ b/Resolver/src/org/icpc/tools/resolver/Resolver.java
@@ -216,7 +216,7 @@ public class Resolver {
 				p = r.problemList[i % r.problemList.length];
 			if (r.problemIdList != null && r.problemIdList.length > 0)
 				pId = r.problemIdList[i % r.problemIdList.length];
-			r.init(steps, g, p, pId, i, resolveInfo.getSingleStepRow(), resolveInfo.getSpeedFactor());
+			r.init(steps, g, p, pId, i, resolveInfo);
 
 			Trace.trace(Trace.INFO, "Resolution steps:");
 			for (ResolutionStep step : steps)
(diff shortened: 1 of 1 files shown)


deboer-tim (approved):
Will fix the immediate problem, but:
- Why not directly pass in the resolveInfo now? Would be simpler code and if we (e.g.) provided a way to set initial scroll speed we wouldn't have to keep adding params.
- The issue now points out the fast/speed inconsistency too, so we should either fix that (the docs, not the code) or remove the 'fixes',

nickygerritsen:
> Will fix the immediate problem, but:
> 
> * Why not directly pass in the resolveInfo now? Would be simpler code and if we (e.g.) provided a way to set initial scroll speed we wouldn't have to keep adding params.
> * The issue now points out the fast/speed inconsistency too, so we should either fix that (the docs, not the code) or remove the 'fixes',

Do you think nothing will break fi we pass the whole info? I can do that

deboer-tim:
> > Will fix the immediate problem, but:
> > 
> > * Why not directly pass in the resolveInfo now? Would be simpler code and if we (e.g.) provided a way to set initial scroll speed we wouldn't have to keep adding params.
> > * The issue now points out the fast/speed inconsistency too, so we should either fix that (the docs, not the code) or remove the 'fixes',
> 
> Do you think nothing will break fi we pass the whole info? I can do that

Yes, that was the original intent - let the command line args customize one ResolveInfo and set it, instead of having multiple vars, passing them along, and then creating the ResolveInfo later on.

nickygerritsen:
> > > Will fix the immediate problem, but:
> 
> > > 
> 
> > > * Why not directly pass in the resolveInfo now? Would be simpler code and if we (e.g.) provided a way to set initial scroll speed we wouldn't have to keep adding params.
> 
> > > * The issue now points out the fast/speed inconsistency too, so we should either fix that (the docs, not the code) or remove the 'fixes',
> 
> > 
> 
> > Do you think nothing will break fi we pass the whole info? I can do that
> 
> 
> 
> Yes, that was the original intent - let the command line args customize one ResolveInfo and set it, instead of having multiple vars, passing them along, and then creating the ResolveInfo later on.

Ok I will change this and test it

PR #997, opened 2024-08-19
Implement new honors awards

### LINKED ISSUE #994: Add new award types to award generator
- `B` can be dropped (but let's leave it for now?)
- From the SysOps agenda: Teams competing in the ICPC World Finals are eligible for Championship Honors awards (defined as follows):
  - Highest Honors: All medalist teams together with all teams solving the same number of problems as the minimum number of problems solved by any medalist team.
  - High Honors: All teams solving one fewer problem than the minimum number of problems solved by any medalist team.
  - Honors: All teams not receiving Highest Honors or High Honors solving the same number of problems or more than the median scoring team.
  - Honorable Mention: All teams solving fewer than the number of problems solved by the median scoring team.

Honorable Mention is as it is before. The other three groups are new. Note that any of these list can be empty if there are no teams within that group.

nickygerritsen:
Update the old honorable mention award template to support both percentiles and 'number of solved problems compared to the lowest medalist'.

Fixes #994.

Add some special cases:
* The highest honors should not be displayed as a list or anything
* Any list that overlaps or is just below the medalists, should be displayed before resolving medalists

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/IAward.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/IAward.java
@@ -52,7 +52,7 @@ public interface IAward extends IContestObject {
 	AwardType GROUP_HIGHLIGHT = new AwardType("Group Highlight", "group-highlight-.*");
 	AwardType SOLVED = new AwardType("Solved", "solved-.*");
 	AwardType TOP = new AwardType("Top", "top-.*");
-	AwardType HONORS = new AwardType("Honors", "honors-.*");
+	AwardType HONORS = new AwardType("Honors", "(honors-.*|.*-honors|honors)");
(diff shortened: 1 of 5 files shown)


nickygerritsen:
@deboer-tim  There is at least one bug: if the 13th team is a regional winner, it will show the highest honors before the regional winner, while it should be the other way around.  This happens because ResolverLogic has:

```java
				ListAwardStep step = getListAward(currentRow, false);
				if (step != null) {
```

before

```java
				if (doneWithRow) {
					// we've finished resolving a team, check if they have any awards
```

But that is probably because of some reason which makes it that we can't just swap that?

deboer-tim (approved):

deboer-tim:
> But that is probably because of some reason which makes it that we can't just swap that?

Maybe just priority of awards? I'd have to take more time to look at it, but should be an easy enough fix either way.

PR #1028, opened 2024-09-15
Add composite presentations in a single dialog and enable editing

zalenski:
- Combines the composite presentation setup dialog and its save dialog.
- Enables editing composite presentations, and duplicating them by choosing a new name

Screenshot 2025-08-30 at 15 57 55


vmcj (approved):
But not an expert on the code..,

There is a lot of code which should be cleaned up though. Probably best to do in next PR.

PR #1041, opened 2024-09-16
Update firstsolutions.txt from WF46 and WF47

zalenski:
From "First to solve" in https://icpc.global/worldfinals/results


deboer-tim:
Nicky already did this?

zalenski:
yes, forgot to fetch, and this was already done in #1038!

PR #1080, opened 2024-12-16
Drop support for 2020-03 spec

deboer-tim:
Removes support for the 2020-03 spec (event feed with 'op'erations) since it hasn't been used in years.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/feed/NDJSONFeedWriter.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/feed/NDJSONFeedWriter.java
@@ -12,7 +12,6 @@ import org.icpc.tools.contest.model.internal.Deletion;
 public class NDJSONFeedWriter {
 	protected PrintWriter pw;
 	protected JSONEncoder je;
-	protected boolean isOldFeed = "true".equals(System.getProperty("ICPC_OLD_FEED"));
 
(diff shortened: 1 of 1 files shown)


nickygerritsen:
Isn't this what PC^2 still uses?

deboer-tim:
I renamed the title to be more accurate: this removes the undocumented ICPC_OLD_FEED support from the CDS for _outputting_ in the 2020-03 event feed format, i.e. if you have clients that still require 2020-03 you would need to keep using the stable 2.5.x version. It does not change what's supported by any of the ICPC tools clients.

i.e. if you want to use a CDS _and_ have 4+ year old clients _and_ were using the undocumented ICPC_OLD_FEED property to support them, you'd need to stick with 2.5.x. 😄 The ICPC tools clients from 2.4 or 2.5 releases can still read either format, so they're unaffected by this change.

nickygerritsen (approved):

clevengr:
Indeed @nickygerritsen is correct; we still support this specification in PC2.  However, since we don't _read_ the CDS output event feed (and I have trouble imaging an application where a client using PC2 would attempt to do so), I don't think we have any reason to object to this.

johnbrvc:
I agree. PC2 can read both the 2020-03 and 2022-07 event feeds and end points.

So if 2020-03 generation is removed from the cds it should not affect us.

PR #1102, opened 2025-04-13
Move some duplication to its own functions

vmcj:
The last 3 commits should be squashed but this makes reviewing a bit easier.


nickygerritsen (approved):

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
```diff
@@ -187,6 +199,57 @@ public void generateTeamPhotos() {
 		generateImages("teams", "photo", "jpg", spec);
 	}
 
+	public static int shrinkDirectionRight(int left, int w, int h, int[] pix, boolean horizontal) throws Exception {
```
deboer-tim, line 202: These are more start/end than right/left since they work in two directions?

deboer-tim (approved):

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
vmcj: Yes, that's a better name.

PR #1108, opened 2025-05-02
First pass to read SVG files

vmcj:
The ImagesGenerator was not able to read SVG files (yet). There was already some code to handle reading and converting `svg`s to preliminary `png`s.

~~I'm currently a bit stuck though as I can either do:~~
~~- the trivial thing is to duplicate the code and alter that to my needs,~~
~~- Alter the ContestObject and use that as a utility class to handle the reading/writing~~

~~As both are dirty I suspect there is a simpler fix which I don't see at the moment, so some guidance/ideas would be helpful.~~

Went with option 2 which still looks like a hack but might be the best option for now.


deboer-tim (approved):
ContestObject isn't a great place to add the static utility function, should probably be on a utility class and used by both places. Fine for now though to get the code in though, can refactored later.

Issue #1076, opened 2024-11-24
Integrate Sentry SDK


nickygerritsen:
It would be nice if we can integrate the [Sentry SDK](https://docs.sentry.io/platforms/java/) and set the DSN with an env var / config.

Then we can configure Sentry (or Glitchtip for offline contests) if we want it and get nice errors reported automatically.

deboer-tim:
I don't have any concern with this, but since I've never used Sentry it's hard to understand the value it provides / how it would be used / how it should be integrated. We need to do this in a way that doesn't interfere with people who aren't using Sentry - and allow that someone else could come along and want to integrate some other tool next year.

It's a little odd that Sentry doesn't list microprofile or vanilla Java in their support list, but assume that's mostly an oversight as you appear to be able to just use the jar.

Finding every place where we do a response.sendError() or printStackTrace() and adding a call to Sentry.captureException(e) is a huge task, would easily get out of sync, and would get even worse if someone else wanted to integrate another tool. If the intention is only to add it in some places we should decide that up front, but if it is generally 'everywhere' we should put the captureException() in Trace, and come up with a similar pattern for HTTP responses (i.e. move all calls to sendError() to one place, then add Sentry there).

PR #1112, opened 2025-05-18
Add Sentry logging for client tools

vmcj:
We use this often for DOMjudge so it helps to detect some of the raised issues automatically.

I didn't update the eclipse part as I don't know (yet) how to set that one up. I think this can be extended to send the error to Sentry each time that the Trace has a level of `ERROR` and merging the `Trace` & `Sentry` calls in a custom wrapper. Before I put in that time I would like to know if this is even something which is wanted.

### DIFF
--- a/ContestUtil/ContestUtil.iml
+++ b/ContestUtil/ContestUtil.iml
@@ -8,5 +8,15 @@
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
     <orderEntry type="module" module-name="ContestModel" />
+    <orderEntry type="module-library">
+      <library>
+        <CLASSES>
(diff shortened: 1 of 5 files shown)


nickygerritsen (approved):
I like sentry so I'm all for it. We should find a way to have the properties file not be hardcoded in git maybe though? I think we could use an env var / system property

vmcj:
> I like sentry so I'm all for it. We should find a way to have the properties file not be hardcoded in git maybe though? I think we could use an env var / system property

My idea was to put it in the .gitignore after this is merged, passing an ENV does also work but makes it harder to set the other properties. So a combination is probably needed.

vmcj:
Linked to: https://github.com/icpctools/icpctools/issues/1076

PR #1113, opened 2025-05-18
Add Sentry support for the CDS

vmcj:
I think this one should be merged before: https://github.com/icpctools/icpctools/pull/1112 as it already implements some of the improvements for that one.

I choose to also raise errors when another system has issues, mostly because those events are important in a contest but might go unnoticed. If this is accepted I can squash the commits but I didn't want to lose progress if changes are requested.

### DIFF
--- a/CDS/config/wlp/cds/server.xml
+++ b/CDS/config/wlp/cds/server.xml
@@ -17,6 +17,7 @@
     <httpSession invalidationTimeout="8h" idReuse="true"/>
 
     <keyStore id="defaultKeyStore" password="{xor}FhwPHAswMDMs" />
+    <!--<ssl id="defaultSSLConfig" keyStoreRef="defaultKeyStore" trustDefaultCerts="true" />-->
 
     <jndiEntry jndiName="icpc.cds.config" value="${server.config.dir}/config"/>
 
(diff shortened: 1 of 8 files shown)


vmcj:
Linked to: https://github.com/icpctools/icpctools/issues/1076

CDS/config/wlp/cds/server.xml
```diff
@@ -17,6 +17,7 @@
     <httpSession invalidationTimeout="8h" idReuse="true"/>
 
     <keyStore id="defaultKeyStore" password="{xor}FhwPHAswMDMs" />
+    <!--<ssl id="defaultSSLConfig" keyStoreRef="defaultKeyStore" trustDefaultCerts="true" />-->
```
nickygerritsen, line 20: What does this do and why can't we enable it by default?

nickygerritsen (approved):

CDS/config/wlp/cds/server.xml
vmcj: It makes that the JVM/CDS server trusts the certificate of sentry.io. Normally we don't have any certificates in the store so there was an SSL error. If people don't use sentry I don't see a reason to insert that many certificates at once. I think it's fine to default change this (as the machine already trusts those certs) but wanted to limit the amount of new features as much as possible.

PR #1115, opened 2025-05-20
Fix svg png same url

vmcj:
This should fix the the `/organizations/id/logo.{svg,png}` issue. Only opening this to make sure no one else works on this in parallel. Did not test with the presentation client yet if it works out of the box.

- The URLs are resolvable and provide the right file.
- The warning about the re-used URL would run on every check for directory changes. This is a bit verbose as this shouldn't happen in practice I think a bit of verbosity is less of an issue (although ugly).
- There is a remark that the current code wouldn't handle a similar case, I'm not sure if its besides similar really this case so I left the `TODO` for now.
- I suspect the `ext` can be added in a better way, this was the result of a long flight of debugging (and not much changes in the end) but a second pass will be needed.

### DIFF
--- a/BalloonUtility/BalloonUtility.iml
+++ b/BalloonUtility/BalloonUtility.iml
@@ -12,6 +12,16 @@
     <orderEntry type="module-library">
       <library>
         <CLASSES>
+          <root url="file://$MODULE_DIR$/../ContestModel/lib" />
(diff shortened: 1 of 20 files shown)


deboer-tim:
We merged #1116 in May. Can we close this draft, or do you still want to change something?

PR #1116, opened 2025-05-22
fix: support file references with same name but different extensions

### LINKED ISSUE #1111: CDS serving default logo with multiple mimetypes but same href
In preparation for NAC2025 we have the following logo's
/organization/<oid>/logo.svg
/organization/<oid>/logo.png
/organization/<oid>/logo.AxB.png

In the eventfeed this is represented as (see the duplicated href with different mimetypes)

```
{"id":"187","icpc_id":"187","logo":[
...,
{"href":"contests/nac25practice/organizations/187/logo","filename":"logo.png","mime":"image/png","width":9887,"height":2787},
{"href":"contests/nac25practice/organizations/187/logo","filename":"logo.svg","mime":"image/svg+xml","width":95,"height":27}]},
...]}
```

The logo.svg & logo.png depict the same file but in different formats, most likely the CDS should generate a different `href` with for example the extension also in the URL and serve the stem only as an redirect to 1 of the 2 for BC.

Image
Image
Image

deboer-tim:
When there are two files with the same name but different extensions, use a different href to refer to them.

This can happen when the folder contains logo.png and logo.svg, or any other resource where there are two or more supported file extensions.

Fixes #1111.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/feed/DiskContestSource.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/feed/DiskContestSource.java
@@ -1092,6 +1092,7 @@ public class DiskContestSource extends ContestSource {
 			folder = new File(root, pattern.folder);
 
 		FileReferenceList refList = new FileReferenceList();
+		List<String> hrefs = new ArrayList<>();
 		for (String ext : pattern.extensions) {
 			File[] files = folder.listFiles(
 					(dir, name) -> (name.toLowerCase().startsWith(pattern.name) && name.toLowerCase().endsWith("." + ext)));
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

ContestModel/src/org/icpc/tools/contest/model/feed/DiskContestSource.java
```diff
@@ -1101,9 +1102,15 @@ private FileReferenceList getFilesWithPattern(FilePattern pattern) {
 					String diff = file.getName();
 					diff = diff.substring(pattern.name.length(), diff.length() - ext.length() - 1);
 					FileReference ref = getMetadata(pattern.url + diff, file);
-					if (ref != null)
+					if (ref != null) {
+						// update the href if it is already in use (e.g. for logo.svg and logo.png)
+						int count = 2;
```
nickygerritsen, line 1107: We should increase count, but isn't #1115 better?

deboer-tim: Oh, I missed the change to line 1104 and thought the draft status meant it wasn't quite done. Yeah, that works too, but I think we should pick an option that can't have conflicts and hence doesn't need the warning output, plus remove the TODO.

I had leaned toward not including the file extension in the href originally so was using the counter instead. I think it would probably be ok to just use the real filename too instead of either of these changes, just less pretty hrefs.

PR #1121, opened 2025-05-24
Show runs for spectators as per new access policy

nickygerritsen:


vmcj:
I did not check fully, but isn't it safer to duplicate for now (so we know that analyst keeps on working) and make the proper PR after NAC?

nickygerritsen:
Analyst directly extends this class so duplicating is a noop

vmcj (approved):

PR #1126, opened 2025-05-26
According to the access policy, the public CAN get submission language

nickygerritsen:
This should fix the language presentations


deboer-tim:
PublicContest really does mean public (the previous definition) and includes teams. Is it intended that teams can see other team's submission language?

vmcj:
> Yes, it's expected that the language stats presentation no longer works (unless you're in sysops/admin) because you won't have access to the data. I'm asking if it's confirmed that spectators should be able to see it again, and if it was intentional to apply to the contest floor where teams can see.

I know John wanted the presentation to be there on the floor, but I assume he was at that moment not aware that the policy doesn't allow this.

nickygerritsen:
I think we should close this and @clevengr should accept that this means the language stats presentation is not useful for teams.

PR #1149, opened 2025-07-18
Fix aspect ratio

vmcj:
In case someone provides logo's which are far from squarish the aspect ratio gets lost as the logo is stretched. As I'm not 100% how far we want to go we should just pick the right commits and squash those as the solution.

commit: Move image size to CSS (shows the problem):
image

commit: Keep original aspect ratio
image

commit: Center the logo over the cell
image
For the last one I think it only works when all cells get centered like this. Maybe there should just be a maximum on the lenght of the display name of the team to get a more consistent scoreboard.

### DIFF
--- a/CDS/WebContent/WEB-INF/jsps/scoreboard.jsp
+++ b/CDS/WebContent/WEB-INF/jsps/scoreboard.jsp
@@ -53,7 +53,7 @@
 <th class="text-center"><span class="badge" style="background-color:{{rgb}}; width:25px; border:1px solid {{border}}"><font color={{fg}}>{{label}}</font></span></th>
 </script>
 <script type="text/html" id="row-start">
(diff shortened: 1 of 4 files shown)


deboer-tim:
FWIW I haven't put a huge amount of effort into the CDS scoreboard as we never tell anyone about it so no real 'users'. I'm good with fixes/improvement, but would put more effort into something that would have external users, has a newer technical base, and potentially could be reused by other tools (e.g. maybe scoreboard in the team view I've been slowly building). Re the commits:

commit: Move image size to CSS
- I'm always against anything that'll stretch logos. :)

commit: Keep original aspect ratio
- +1

commit: Center the logo over the cell
- Agree we could limit # of lines and centre everything, but not sure I'd like it / prefer to keep PR to 1 issue.

vmcj:
Removed the centering commit, removed the commit of code cleanup.

deboer-tim (approved):

PR #1164, opened 2025-08-06
Support for multiple teams and groups in clarification destinations

deboer-tim:
Adds support for the draft Contest API ability to send clarifications to multiple teams and/or groups.

This updates the contest model and all client-side support to support multiple teams and groups. The CDS, presentations, and all other tools will work when pointing to a 2023-06 _or_ draft spec server.

By default the CDS will continue to serve the 2023-06 contest API though; if you want the new event feed you need to set the ICPC_CONTEST_API=draft system property as with all other draft spec changes.

CDS web pages will need to be updated as well, that will be a separate PR.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
+++ b/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
@@ -747,9 +747,20 @@ public class ConfiguredContest {
 			ITeam team = contest.getTeamById(clar.getFromTeamId());
 			if (contest.isTeamHidden(team))
 				return null;
-			team = contest.getTeamById(clar.getToTeamId());
-			if (contest.isTeamHidden(team))
-				return null;
+			if (clar.getToTeamIds() != null) {
+				for (String teamId : clar.getToTeamIds()) {
+					team = contest.getTeamById(teamId);
+					if (contest.isTeamHidden(team))
+						return null;
+				}
+			}
+			if (clar.getToGroupIds() != null) {
+				for (String groupId : clar.getToGroupIds()) {
+					IGroup group = contest.getGroupById(groupId);
+					if (group != null && group.isHidden())
+						return null;
+				}
+			}
 		}
 
 		return obj;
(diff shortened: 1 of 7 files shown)


ContestModel/src/org/icpc/tools/contest/model/internal/Contest.java
```diff
@@ -1809,7 +1809,10 @@ public void removeHiddenTeams() {
 				remove.add(clar);
 				foundClar = true;
 			}
-			teamId = clar.getToTeamId();
+
+			// For now, assume that clarifications to hidden teams are only
+			// sent to hidden teams. (i.e. no complex groups or hidden+non-hidden teams)
```
vmcj, line 1814: I think this should also become part of the spec, so it's easier for someone to implement the spec and not have this edge case.

vmcj (approved):
LGTM but feel free to wait for a review from someone more experienced with the code.

CDS/src/org/icpc/tools/cds/ConfiguredContest.java
```diff
@@ -747,9 +747,20 @@ protected IContestObject filterHidden(IContestObject obj) {
 			ITeam team = contest.getTeamById(clar.getFromTeamId());
 			if (contest.isTeamHidden(team))
 				return null;
-			team = contest.getTeamById(clar.getToTeamId());
-			if (contest.isTeamHidden(team))
-				return null;
+			if (clar.getToTeamIds() != null) {
```
nickygerritsen, line 750: I wonder if it would be better to only hide the clarification if **all** teams and groups in it are hidden and if not, only hide the team / group IDs of hidden ones? It does complicate stuff though.

ContestModel/src/org/icpc/tools/contest/model/internal/Contest.java
nickygerritsen: See also my remark above

PR #1166, opened 2025-08-07
URL() deprecation

deboer-tim:
The constructor 'new URL(x)' has been deprecated in Java 20 because it never correctly handled encoding & decoding. The simplest switch is just to use 'new URI(x).toURL()': https://stackoverflow.com/questions/75966165/how-to-replace-the-deprecated-url-constructors-in-java-20

The bigger ripple is the exception for a bad url isn't MalformedURLException (an IOException) anymore. Switched to using a generic Exception instead.


vmcj (approved):
I think I had a similar branch somewhere with the same solution.

PR #1167, opened 2025-08-11
Configurable video handlers

deboer-tim:
The CDS has a default video handler (MPEG-TS), which can already be changed via the ICPC_VIDEO system property before starting the CDS. e.g. Using 'ICPC_VIDEO=hls' would change all video streaming to HLS.

This change provides a bit more configuration:
- You can specify which handler to use / override the default using '<name>|url' directly in the CDS config.
- You can use ',' to separate multiple handlers - e.g. to provide both MPEG-TS and HLS for the same team stream.

For example, the config:
`<video desktop="hls|http://localhost:8081,ogg|http://localhost:8082" webcam="http://localhost:9081"/>`

Would set up _both_ HLS and OGG streaming for desktop streaming, and MPEG-TS for the webcam:
```
Video reservation for DESKTOP 70 (HLS) at http://localhost:8081 on stream 417
Video reservation for DESKTOP 70 (OGG) at http://localhost:8082 on stream 418
Video reservation for WEBCAM 70 (MPEG) at http://localhost:9081 on stream 419
```

And the contest feed would look like this:
```
{"id":"70",
 "desktop":[
   {"href":"https://10.0.0.198/stream/417","mime":"application/vnd.apple.mpegurl"},
   {"href":"https://10.0.0.198/stream/418","mime":"video/ogg"}],
 "webcam":[
   {"href":"https://10.0.0.198/stream/419","mime":"video/m2ts"}]...}
```
i.e. the desktop shows two streams and the webcam one, each with correct mime types.

Notes:
- The main reason for this change was to remove most of the remaining hard-coding/ defaulting from the video streaming pipeline.
- The secondary point was to make the CDS configurable instead of using system properties. I don't actually think someone would use the config above, it just shows what's possible. e.g. maybe we want to continue to stream MPEG-TS while we also try another format.
- I used ',' and '|' because I needed characters that don't normally exist in URLs, if anyone has a better suggestion or better config I can switch (or we can do it later).

### DIFF
--- a/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
+++ b/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
@@ -374,11 +374,11 @@ public class ConfiguredContest {
 		map.put(type, in);
 
 		String name = type.name() + " " + teamId;
-		StringTokenizer st = new StringTokenizer(urlPattern, ",");
+		StringTokenizer st = new StringTokenizer(urlPattern, "|");
 		while (st.hasMoreTokens()) {
 			String urlPattern2 = st.nextToken();
 			String handler = null;
-			int index = urlPattern.indexOf('|');
+			int index = urlPattern.indexOf('>');
 			if (index > 0) {
 				handler = urlPattern2.substring(0, index);
 				urlPattern2 = urlPattern2.substring(index + 1);


vmcj (approved):
I think `,` is valid for URLs so I propose to pick something else (see: https://stackoverflow.com/a/7109208)  ` seems to be fine though.

Also as a sidenote, I think `application/vnd.apple.mpegurl` is not valid according to the CLICS spec. If it is I suggest to rephrase it.

deboer-tim:
Rebased on main and added a new commit to switch the separators to `desktop="hls>http://localhost:8081|ogg>http://localhost:8082"`, let me know if that's ok for now or you have a better suggestion.

PR #1169, opened 2025-08-19
Define codestyle for editors

vmcj:
My editor has a strong preference for 4-space indentation. This should take precedence and also nudge the editors of other contributors IIRC.

This does not work with Eclipse without a plugin though: https://editorconfig.org/ but as Tim has 4x more commits as the next most contributor I would say that his editor already defines the code standard.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = false
+trim_trailing_whitespace = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset
+[*.{js,java}]
+charset = utf-8
+indent_style = tab


vmcj:
> On no, we've reached the 'tabs vs spaces' debate. :)
> 
> I'm not opposed to .editorconfig, but adding this w/o doing something about the existing Eclipse org.eclipse.jdt.core.prefs files creates a 'formatting conflict'. If I delete those we have to document for Eclipse users to install an extension, which is ok, but seems a bit backwards.

I was not aware of those eclipse*prefs, do those still work when you translate them to the `xml` files? It seems that IDEA (after installing a plugin) can read those which would be a good intermediate solution.

> 
> For node projects I work on we use a command line tool like prettier (now biome) to do all formatting, so that you can easily clone, edit, and format regardless of IDE. We often use husky (git hooks) to enforce it on commit. This way the entire codebase stays consistently formatted, even random commits done using notepad or from external contributors.
> 
I like this but I'm not sure if every contributor knows of git hooks. Looking at the current list of external contributors I think this is more a theoretical discussion though. 

> I'm not picky about which formatter to use (there are a few [1]) and don't need to do the git hooks bit, but maybe we can pick one and configure the repo so that you can easily clone and run it _outside_ of an IDE? Then I'd be happy to delete the Eclipse rules (or could configure it to run the formatter on save). The main difference to me is setting it up so that it's editor agnostic (even if some editors have built-in support) vs trying to synchronize tools in IDEs.
> 

I think the idea of `.editorconfig` was to become agnostic but it's probably one of those new standards to replace all other standards resulting in `x+` standards. I have read the reddit article and don't see a clear winner. Let's discuss onsite if this is something the project needs. I'll convert this to draft as reminder to discuss it.

PR #1172, opened 2025-08-21
Specify which registry to trust

vmcj:
The Docker binary defaults to docker.io as registry. Podman as alternative doesn't and might ship with no default registry so can't build this container.

By explicit specifying the registry the risk for inconsistent builds is also decreased in case the GHA ever encounters a similar named container on the GHCR and would not use docker.


deboer-tim (approved):
This is just locking down what's already there, so approving.

OpenJDK images have been deprecated [1], we should probably switch to Temurin [2] soon.

[1] https://hub.docker.com/_/openjdk
[2] https://hub.docker.com/_/eclipse-temurin

PR #1173, opened 2025-08-21
World map logo ratio

### LINKED ISSUE #1171: World map: long institution name causes logo to compress
When running the presentation client and showing the Team Intro, we notice that sometimes, the text at the bottom is causing the university logo to get out of ratio. This only happens with very long university names, see below:

Image

vmcj:
Closes #1171,

I'm not 100% if the logo is now placed in the same location or if the `x` should be calculated differently but on first glance this looks like a fix.

I've added 2 commits, for both possible solutions.

I like the 2nd for the consistency
The first has the benefit of giving more space for the name which might be worth it as the logo is already on screen at that moment.

image
image

### DIFF
--- a/PresContest/src/org/icpc/tools/presentation/contest/internal/presentations/map/TeamIntroPresentation.java
+++ b/PresContest/src/org/icpc/tools/presentation/contest/internal/presentations/map/TeamIntroPresentation.java
@@ -430,6 +430,10 @@ public class TeamIntroPresentation extends AbstractICPCPresentation {
 		}
 
 		text.addString(pos.label);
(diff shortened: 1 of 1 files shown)


deboer-tim:
For the first image, isn't it cutting off the logo to the left _and_ the name on the right?

The second doesn't look centred any more, and isn't it also cutting off the long name at the right?

The fix I was thinking of was doing a bit more complicated scaling to set just the logo part as 'non-shrinkable' - but that would need to be done carefully as there can be other images (e.g. emojis embedded in the text) that you'd still want to shrink. Not sure it's worth doing that way.

deboer-tim (approved):

Issue #976, opened 2024-05-25
Make coachview banner part have a dark background


nickygerritsen:
It has a light background currently, which makes it harder to have one banner for both presentations and the coach view

deboer-tim:
Closing this as the long-term plan for coach view is to replace it with team view - which already has support for light/dark mode and light/dark logos/banner.

PR #1177, opened 2025-08-21
Use darkbackground in all ICPC tools

vmcj:
During NAC we had to change the banner to make sure it looks nice on Coachview. There was already an issue (https://github.com/icpctools/icpctools/issues/976) to change the color to Black.

Screenshot from 2025-08-21 22-54-08

### DIFF
--- a/CoachView/src/org/icpc/tools/coachview/CoachView.java
+++ b/CoachView/src/org/icpc/tools/coachview/CoachView.java
@@ -82,6 +82,9 @@ public class CoachView extends Panel {
 
 	private static Map<String, Locale> localeMap;
 
+	private static boolean lightMode = false;
+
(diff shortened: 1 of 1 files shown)


vmcj:
> We've also had banners that only work for light mode, and I'm sure someone is old school and will want that back. (plus of course those who would prefer we pick up the system default...)
> 
> IMHO if we're going to do this, at a minimum we should support the same `--light` option as the presentations (or the reverse). (Or get HLS and the team-view over the line to replace coachview entirely 😄)

My idea was to look at the `--light` next (but only for the presentations)

As you were already working on a replacement for CoachView I didn't want to spend that much time on this. I'll see how much work it is to just implement `light` mode.

PR #1180, opened 2025-08-25
Default presentation for presentation client

### LINKED ISSUE #955: p100 should default to teamdesktop
I don't care if this is implemented as a default if "--name team" arg is used on the org.icpc.tools.presentation.contest.internal.ClientLauncher

Or if it's a something in cdsConfig.xml for that server. 

This is addressing an issue we had at Egypt and NAC2023 where the p100 comes up with the tools branding.   And if the CDS is restarted and a p100 dies and is restarted it comes up with tools branding.   Either way requires a CDS click of teamdesktop to fix.

deboer-tim:
Adds an option to set a default presentation for connected presentation clients. This presentation is set before trying to connect to the CDS. If the CDS doesn't set any presentation then this one is shown; otherwise whatever the CDS says takes over.

This avoids the standard ICPC Tools splash page, and allows you to have a default w/o waiting for clients to connect and then setting it in the pres admin.

Fixes #955.

### DIFF
--- a/PresContest/src/org/icpc/tools/presentation/contest/internal/ClientLauncher.java
+++ b/PresContest/src/org/icpc/tools/presentation/contest/internal/ClientLauncher.java
@@ -10,9 +10,13 @@ import org.icpc.tools.contest.model.feed.RESTContestSource;
 import org.icpc.tools.contest.model.util.ArgumentParser;
 import org.icpc.tools.contest.model.util.ArgumentParser.OptionParser;
 import org.icpc.tools.contest.model.util.TeamDisplay;
+import org.icpc.tools.presentation.contest.internal.standalone.PresentationHelper;
+import org.icpc.tools.presentation.contest.internal.standalone.StandaloneLauncher;
 import org.icpc.tools.presentation.contest.internal.standalone.TeamUtil;
 import org.icpc.tools.presentation.core.DisplayConfig;
+import org.icpc.tools.presentation.core.Presentation;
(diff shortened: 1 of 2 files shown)


PresContest/src/org/icpc/tools/presentation/contest/internal/ClientLauncher.java
```diff
@@ -110,8 +123,39 @@ public void showHelp() {
...
+		if (presStr[0] != null) {
+			Trace.trace(Trace.INFO, "Default presentation: " + presStr[0]);
+			List<PresentationInfo> presentations = PresentationHelper.getPresentations();
+			if (presentations.isEmpty()) {
+				Trace.trace(Trace.USER, "No presentations found");
+				System.exit(0);
+			}
+
+			PresentationInfo pi = StandaloneLauncher.findPresentation(presentations, presStr[0]);
+			if (pi == null)
```
nickygerritsen, line 137: Maybe also add an trace here that the presentation doesn't exist?

nickygerritsen (approved):

PresContest/src/org/icpc/tools/presentation/contest/internal/ClientLauncher.java
deboer-tim: findPresentation() already traces and does a system exit if the presentation can't be found or matched so this code is unnecessary and could be removed, only kept it since it matches what the standalone client had.

PR #1181, opened 2025-08-25
Fix lightmode some presentations

vmcj:
I think I need some help to check for the lightMode in a proper way. Best would be an autowire but I don't have enough experience with Java on how to do this properly. Is there a piece of code which I can look at to use as example?

On itself all are improvements IMO as the the presentations didn't work in lightmode. I checked with a client also in normal/dark mode and it seems everything stayed the same there.

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/FloorMap.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/FloorMap.java
@@ -34,12 +34,14 @@ public class FloorMap {
 
 	// default colors for printing
 	public static class FloorColors {
+		protected boolean lightMode = false;
+
 		public Color getTeamAreaFillColor() {
 			return null;
 		}
 
 		public Color getTeamAreaOutlineColor() {
-			return Color.LIGHT_GRAY;
+			return lightMode ? Color.DARK_GRAY : Color.LIGHT_GRAY;
 		}
 
 		public Color getSeatFillColor() {
@@ -47,7 +49,7 @@ public class FloorMap {
 		}
 
 		public Color getSeatOutlineColor() {
-			return Color.LIGHT_GRAY;
+			return lightMode ? Color.DARK_GRAY : Color.LIGHT_GRAY;
 		}
 
 		public BufferedImage getTeamLogo(String id) {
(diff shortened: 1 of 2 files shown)


deboer-tim:
> I think I need some help to check for the lightMode in a proper way. Best would be an autowire but I don't have enough experience with Java on how to do this properly. Is there a piece of code which I can look at to use as example?

Not sure I understand? We made light mode a basic property on presentations and (intentionally) didn't initially try to support it in all the presentations. The kind of changes I see here are expected, and I'd suggest just PRing it every couple presentations to keep things moving.

> On itself all are improvements IMO as the the presentations didn't work in lightmode. I checked with a client also in normal/dark mode and it seems everything stayed the same there.

Sounds good.

I did notice an unrelated change to using Taskbar in PresentationWindowImpl. I kind of assumed every typical Windows/Mac/Linux system had a taskbar and app icons - are you using one that doesn't have a taskbar? If so, maybe we need that same change everywhere I touched taskbar (but in a separate PR of course, I could do it).

vmcj:
> I did notice an unrelated change to using Taskbar in PresentationWindowImpl. I kind of assumed every typical Windows/Mac/Linux system had a taskbar and app icons - are you using one that doesn't have a taskbar? If so, maybe we need that same change everywhere I touched taskbar (but in a separate PR of course, I could do it).

I'll check around a bit more what is actually failing, this was the fastest fix. I do get the App icon in my dock when the app is started though which implies that something is being set.

> > I think I need some help to check for the lightMode in a proper way. Best would be an autowire but I don't have enough experience with Java on how to do this properly. Is there a piece of code which I can look at to use as example?
> 
> Not sure I understand? We made light mode a basic property on presentations and (intentionally) didn't initially try to support it in all the presentations. The kind of changes I see here are expected, and I'd suggest just PRing it every couple presentations to keep things moving.

I now need to add a property to TextHelper and set the `lightmode` explicit instead of it being autowired from when the property is set on startup with `--light`. This means that in all presentation we would need to change every `TextHelper` to get the explicit call to `setLightMode` which felt a bit overly verbose. I'm fine with doing it like this for now if we want to make sure no other presentations are unintentionally changed.

PR #1184, opened 2025-08-27
Taskbar guard

### LINKED ISSUE #1183: Taskbar API not working on Linux
During development I had to change the Taskbar handling. We should investigate if this is only a problem on my machine or this is generic for Linux. See some discussion in #1181.

Tested with PopOS 22.04 with both KDE and the PopOS Gnome fork this seems to fail.

image

deboer-tim:
Avoid calling Taskbar.setIconImage() on platforms that don't support it.

Fixes #1183.

### DIFF
--- a/ProblemSet/src/org/icpc/tools/contest/util/problemset/ProblemSetEditor.java
+++ b/ProblemSet/src/org/icpc/tools/contest/util/problemset/ProblemSetEditor.java
@@ -537,11 +537,14 @@ public class ProblemSetEditor {
 		}
 
 		Display.setAppName("Problem Set Editor");
-		try {
-			BufferedImage image = ImageIO.read(ProblemSetEditor.class.getResourceAsStream("/images/problemSetIcon.png"));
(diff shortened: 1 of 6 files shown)


ProblemSet/src/org/icpc/tools/contest/util/problemset/ProblemSetEditor.java
```diff
@@ -537,11 +537,14 @@ public static void main(String[] args) {
...
-			Taskbar.getTaskbar().setIconImage(image);
-		} catch (Exception e) {
-			// could not set icon
+		if (Taskbar.isTaskbarSupported()) {
+			try {
+				BufferedImage image = ImageIO
+						.read(ProblemSetEditor.class.getResourceAsStream("/images/problemSetIcon.png"));
+				Taskbar.getTaskbar().setIconImage(image);
+			} catch (Exception e) {
+				// could not set icon
```
vmcj, line 546: I assume you want to Trace.

vmcj (approved):
Nitpick

PR #1195, opened 2025-08-30
Set up floor generator for systest 1 and remove team 92

deboer-tim:


vmcj (approved):
I'll need to see the output but makes sense.

PR #1196, opened 2025-08-30
Fix balloon printer and update SWT launcher

deboer-tim:
Adding the sentry dependency broke a few of the tool scripts a couple months ago. Instead of the direct fix of adding sentry to all the scripts, I tried to simplify and allow all of them to run against all jars in the lib folder (June 20 / build 1215), but I had forgotten how the SWT launcher works.

This may be going further down the rabbit hole, but it simplifies how the SWT launcher works as well, removes unused (and potentially confusing) manifest files, and fixes the scripts to match.

### DIFF
--- a/BalloonUtility/scripts/balloon.bat
+++ b/BalloonUtility/scripts/balloon.bat
@@ -22,7 +22,7 @@ robocopy "%ROOTDIR%\update" "%ROOTDIR%\" /e /move
 
 :restart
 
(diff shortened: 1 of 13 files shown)


vmcj (approved):
LGTM but maybe another Java expert should also look at it. I do think this is beter though.

vmcj (approved):

PR #1202, opened 2025-08-31
Team intro presentation fails when teams are missing data

deboer-tim:
Team intro throws an NPE when there's a team that isn't in any groups - in our case due to bad data, but we should avoid the NPE.


zalenski (approved):
👍

PR #1236, opened 2025-09-30
Improve mime type handling, especially for video

deboer-tim:
Fixes three things:

1. Move the existing utility functions out of DiskContestSource to create a reusable mime type utility class, to avoid some of the duplication and differences in behaviour.

2. The mime type wasn't being set when downloading reaction videos - but using the new util fixes this.

3. When we started supporting MPEG-TS there was conflicting information about the correct file extension and mime type, and based on google AI and video.js we got it wrong:

```
   MPEG Transport Stream (MPEG 2 transport stream) <-- this is what we actually support
   File extension: .ts
   Mime type: video/mp2t
   188-byte packets

   MPEG-2 Transport Stream <-- variant used by Blu-ray & AVCHD disks
   File extension: .m2ts
   Mime type: video/m2ts
   192-byte packets (4 bytes added to help file seeking)
```

   This switches us to use the file extension and mime type of the first.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/util/HttpHelper.java
+++ b/CDS/src/org/icpc/tools/cds/util/HttpHelper.java
@@ -7,6 +7,7 @@ import java.io.IOException;
 import java.util.Enumeration;
 
 import org.icpc.tools.contest.model.feed.JSONEncoder;
+import org.icpc.tools.contest.model.internal.MimeUtil;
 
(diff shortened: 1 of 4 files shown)


nickygerritsen:
This also means we can drop https://github.com/icpctools/team-view/blob/main/src/lib/ui/Video.svelte#L45-L48

PR #1244, opened 2025-11-03
Take into account tags when generating images

nickygerritsen:
This also fixes an issue when requesting an image with tag `x`: it could return an image without tag `x` but with tag `y`, even if there is also an image without any tags. Now we first check if we have any images without any tag and return that.

### DIFF
--- a/ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
+++ b/ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
@@ -19,8 +19,9 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
-import java.util.regex.Matcher;
+import java.util.Map;
 import java.util.regex.Pattern;
 
 import javax.imageio.IIOImage;
@@ -321,9 +322,9 @@ public class ImagesGenerator {
 						name = org.getActualFormalName();
 
 					File orgFolder = new File(orgRootFolder, org.getId());
-					File imgFile = getDefaultFile(orgFolder, "logo", null);
-					if (imgFile.exists())
-						mod = imgFile.lastModified();
+					Map<String, File> imgFiles = getDefaultFiles(orgFolder, "logo");
+					if (imgFiles.get(null) != null && imgFiles.get(null).exists())
+						mod = imgFiles.get(null).lastModified();
 
 					logoImg = org.getLogoImage(1920, 1080, FileReference.TAG_DARK, true, true);
 				}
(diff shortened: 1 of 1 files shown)


ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
```diff
@@ -427,6 +434,8 @@ private static void generateImageSpec(BufferedImage img, ImageSpec[] spec, File
 	}
 
 	public void generateImages(String objectName, String property, String preferredExtension, ImageSpec[] spec) {
+		String[] tags = {null, FileReference.TAG_LIGHT, FileReference.TAG_DARK};
```
deboer-tim: This would be fine for now (covers light/dark tags), but if someone used a different tag it wouldn't get processed. I'm wondering if we should be a bit more flexible now. e.g. maybe we still pick up poorly-named files if it doesn't follow the spec, but if someone has `logo.green-bg.png` we would interpret `green-bg` as a tag and process it to `logo.green-bg.128x128.png` etc?

ContestModel/src/org/icpc/tools/contest/model/internal/ContestObject.java
```diff
@@ -395,22 +395,26 @@ private static FileReference getBestFileReference(FileReferenceList list, int wi
...
-							found = true;
+						if (tag.equals(t)) {
+							listWithApplicableTag.add(ref);
+							break;
+						}
 					}
-				}
-				if (found) {
-					list3.add(ref);
+				} else {
```
deboer-tim: Two concerns here:
- looking at this change it only looks for the correct tag or else untagged images - but if you're looking for `light` tag and the only available image has a tag `green-bg` we would say there's no image.
- the spec doesn't help things here, because we don't know whether an untagged image really is any better than another tag. I almost think we should understand tag-opposites here, i.e. if you're looking for `light` tag and there are none we would prioritize anything 'not dark' next, then anything else (which means dark tag...). 😆

The second is probably overkill, but I think we probably shouldn't cause the first. Maybe this means we still prefer untagged over any other tag 🤷🏼‍♂️ which may not be useful, but at least it wouldn't fail if another tag is found, or we'd fallback to `dark` if there was no `light` or untagged image.

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
nickygerritsen: If you can come up with a nice way to do this, I'm all for it. As you said I think this is a good start at least.

ContestModel/src/org/icpc/tools/contest/model/internal/ContestObject.java
nickygerritsen: > Two concerns here:
> 
> * looking at this change it only looks for the correct tag or else untagged images - but if you're looking for `light` tag and the only available image has a tag `green-bg` we would say there's no image.

Does it though? If there are no images without a tag, the list is empty and I don't use it but use the default list, so it would return that one, right?

> * the spec doesn't help things here, because we don't know whether an untagged image really is any better than another tag. I almost think we should understand tag-opposites here, i.e. if you're looking for `light` tag and there are none we would prioritize anything 'not dark' next, then anything else (which means dark tag...). 😆

Yeah, the main issue I had is for NWERC I will have `logo.svg` and `logo.dark.svg`. No `logo.light.svg` since DOMjudge doesn't handle tags yet. Currently you will get `logo.dark.svg` when requesting the `light` tag if that logo is the first in the list. My change fixes that, but if you have a better way to do it, I'm up for it.

> 
> The second is probably overkill, but I think we probably shouldn't cause the first. Maybe this means we still prefer untagged over any other tag 🤷🏼‍♂️ which may not be useful, but at least it wouldn't fail if another tag is found, or we'd fallback to `dark` if there was no `light` or untagged image.

ContestUtil/src/org/icpc/tools/contest/util/ImagesGenerator.java
nickygerritsen: Claude came up with something that I think is nice, so I updated the code.

deboer-tim (approved):
I'm good with the changes in ImagesGenerator. We may still need tweaks, but I agree that's generally a big step in the right direction.

I tested with the ContestObject changes, and I was able to reproduce the problem I expected where if all images are tagged we can fail to pick, or pick the 'opposite' tag over another option.

It's arguable whether this is a 'problem', but if you ask w/o specifying a tag we may also pick the 'dark' or 'light' instead of untagged, depending on what order the file refs are in.

So: feel free to merge just the IG or both, I'll have a PR ready to go that will improve these two issues in getBestFileReference.

Side note: I think the last issue highlights the fact that CDS and CCSs should probably sort file refs to help out 'dumb'/old clients, even if that's not required by spec.

PR #1276, opened 2026-02-24
Fix alternating audio in team display presentation

### LINKED ISSUE #1275: Team display presentation plays same gender audio every time.
It seems the team display presentation always plays the same gender. I found two issues, which might explain this:

* The `.female.` file gets assigned both a `male` and `female` tag.
* The `FileReference` constructor doesn't set tags from JSON, which we might want to do,.

nickygerritsen:
I found two issues:

* The disk contest source didn't compare tags of files during merging, which would result in both files being merged into one
* The file reference constructor from JSON didn't set the tags

After fixing both I can successfully alternate audio genders.

Fixes #1275

### DIFF
--- a/ContestModel/src/org/icpc/tools/contest/model/internal/FileReference.java
+++ b/ContestModel/src/org/icpc/tools/contest/model/internal/FileReference.java
@@ -38,6 +38,12 @@ public class FileReference {
 		height = obj.getInt("height");
 		width = obj.getInt("width");
 		filename = obj.getString("filename");
+		Object[] tagsArray = obj.getArray("tags");
+		if (tagsArray != null) {
+			tags = new String[tagsArray.length];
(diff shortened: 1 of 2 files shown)


ContestModel/src/org/icpc/tools/contest/model/internal/FileReference.java
```diff
@@ -48,7 +54,7 @@ public void updateTags() {
 
 		List<String> list = new ArrayList<>(KNOWN_TAGS.length);
 		for (String t : KNOWN_TAGS) {
-			if (filename.contains(t))
+			if (filename.matches(".*\\b" + t + "\\b.*"))
```
deboer-tim: My regex fu is not great, but tags will always be between periods (i.e. `.tag.`), so wouldn't `filename.contains("." + t + ".")` be good enough?

PR #1292, opened 2026-03-21
Prevent resolver from happily running with partially parsed data.

meisterT:
Nowadays, event feeds including a million runs can be very large, and parsing takes some time, especially on very old hardware.

This change
- doubles the timeout, as we ran into the same error now multiple times
- and if it still not successful, bails out instead of "silently" continueing

### DIFF
--- a/Resolver/src/org/icpc/tools/resolver/Resolver.java
+++ b/Resolver/src/org/icpc/tools/resolver/Resolver.java
@@ -485,8 +485,10 @@ public class Resolver {
 				Trace.trace(Trace.INFO, "Test mode active");
 			} else {
 				// not test mode
-				if (!contestSources[con].waitForContest(30000))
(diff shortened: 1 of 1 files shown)


deboer-tim:
Does anyone on this issue have access to a large event feed + old or Windows hardware that they know was failing before (or at least could verify current behaviour)?

I ask because there have been several other PRs that should have improved this, including improve loading time for on-disk contests (#1299), a fix to load timeouts (#1339), a fix for when you're loading an event-feed directly (#1256). Then there are resolving improvements like avoiding loading runs in the resolver (#1303) and even a 40% improvement in doing the actual resolving (#1335).

Net: there have been enough improvements that either this isn't a problem anymore, or overall it's much better and the problem/bottleneck is in an entirely different place. I think we should either verify and close this, or just close it until we have another chance to test.

PR #1293, opened 2026-03-22
Update broken links and other content errors in various READMEs

JoelNiemela:
I noticed that multiple links have become outdated, so I went through all `.md` files I could find to try to fix this.

Most of the changes are one of the following:
  - Add a version number to ccs-specs.icpc.io links. I elected to go with `2026-01`. Let me know if you think it would make more sense to still use `2023-06`.
  - Replace http://pc2.ecs.csus.edu/pc2 with https://pc2ccs.github.io
  - Set a width to the various icon images in READMEs, as the layout of them is currently quite broken.
  
Moreover, I found this email listed in `CDS/README.md`: `cpctools-dev@ecs.csus.edu`. As a lot of links (including `pc2.ecs.csus.edu`) have become outdated, I thought I'd double-check if this email is still monitored? If not, it should probably be removed/replaced.

### DIFF
--- a/BalloonUtility/README.md
+++ b/BalloonUtility/README.md
@@ -1,6 +1,6 @@
 # The ICPC Balloon Utility 
 
-![](docs/balloonIcon.png)
+<img src="docs/balloonIcon.png" width="50">
 
 An ICPC Tool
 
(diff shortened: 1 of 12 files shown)


deboer-tim (approved):
Thanks for the updates!

Issue #1310, opened 2026-05-14
Strict spec versioning


deboer-tim:
Clarifications in 2023-06 have a property:
`to_team_id - ID ?`
in 2026-01 this changed to an array:
`to_team_ids - array of ID ?`

At NAC there was an issue where a CCS was sending both properties but the values didn't match, so 'last one wins' instead of the current spec version being respected.

The contest model has to be strict: checking the version first, and then actively ignoring all properties that aren't from that version.

deboer-tim:
Summary from a brief discussion we had after the last sysops call:
- we need to do a better job of storing api.json in contest config, /results or /other-systems (related to this issue, but not trying to solve here)
- if there's no api.json it's really up to the reader to decide which spec(s) to try
- CDS needs a strict mode - but just as much it needs the ability to override it

Since the meeting I've realized that the last bit is even more complicated: for a contest like NAC the CDS needs the ability to pick up the version from disk, force an override, then check the version from the CCS but not allow override (i.e. override must be per-contest source).

PR #1326, opened 2026-07-16
Strict spec versioning

deboer-tim:
Adds strict spec version parsing where individual properties can be read or ignored based on the version returned by the REST contest's API endpoint, starting with clarifications.

Changes the spec version on ContestAPIHelper to be public and cleans up the API a little.

Caches the spec version when you connect to a REST contest API, and logs if the spec version is unknown. When it is known it's set on the parser so that any contest object can check which version we expect we're reading from.

For now, only Clarifications are updated to switch which version of the to_team_ids property is parsed. We could be super-strict on other properties too (e.g. only parse the expected integer or RELTIME for penalty time) but that's a lot of change for no direct benefit, so this provides the mechanism we could do that but no other changes.

Fixes #1130.


nickygerritsen (approved):

deboer-tim:
Typo: fixes #1310.

PR #1329, opened 2026-08-06
Fix missing judgement events in replay

deboer-tim:
This fixes two issues related to the new CDS contest replay:

1. When replaying a past contest, the next event is found by scanning objects in a list, an event is fired, and then the object is removed from the list. However, because they're removed from the list by object (id) it can remove the wrong instance of an object if anything was out of time order. e.g. if a final judgement was in the list before the start of judgement, it can fire the start judgement, remove the end judgement, and then fire the start again: no final judgement is ever sent. This fixes it to remove from the list based on the index.

2. A contest I tested had an invalid contest freeze time, which caused exceptions when the contest state froze later on. This just adds guards to avoid NPEs.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/util/ReplayContest.java
+++ b/CDS/src/org/icpc/tools/cds/util/ReplayContest.java
@@ -131,17 +131,20 @@ public class ReplayContest extends CDSContest {
 		while (!timedObject.isEmpty() && !stopReplay) {
 			IContestObject nextObj = null;
 			long nextTime = Long.MAX_VALUE;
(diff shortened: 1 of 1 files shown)


johnbrvc (approved):
Looks good.  It seems that if no freeze duration is specified (regardless of whether the one that was specified was bad), it would have gotten an NPE as anyway since it looks like `Info.freezeDuration` is not initialized to anything (`null`).  (Maybe I didn't follow it deep enough, but ...) Anyway, the checks added should address that for sure.

Issue #1270, opened 2026-02-11
Resolver control feed output


deboer-tim:
The resolver currently works by all clients having full access to the feed, each one determining the full resolution flow in advance, and the presenter client sending the step count to the other clients - basically just 'go to step 47'. Each resolver client already knows whether that's a scoreboard, selecting a run, award screen, etc.

I did some work a few years ago to allow the resolver control to expose event feed data as this process happens - e.g. unfreezing individual judgements in the public feed as they are resolved. This only works if the resolving was done on the CDS though (not via the main resolver client).

I want to open this up so that _any_ client connected to the same CDS can see these events _and_ some basic resolver status, e.g. when we show an award screen. To be specific:

- Public clients connected to the CDS should be able to see all resolved runs and the current scoreboard during the entire resolving process, using the regular Contest API updates.
- A new event would indicate what screen the resolver is currently on. Initially this would only cover the main screens and basic information, e.g. scoreboard or award screen.

Net: I should be able to build a web client that would show scoreboard updates and what award is currently being presented, in sync with the resolver.

PR #1338, opened 2026-08-17
Resolver feed undo and award support

deboer-tim:
Changes three things:
1. Undo (removing judgements) wasn't working because to delete something from a contest, you don't remove it - 'you add a deletion event'. Fixed.

2. If you start the resolution process from the CDS admin and there are no awards, it will create default ones (same behaviour as the regular resolver).

3. Added awards to the feed (also with working undo), just like judgements.

Part of #1270.

### DIFF
--- a/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
+++ b/CDS/src/org/icpc/tools/cds/ConfiguredContest.java
@@ -21,6 +21,7 @@ import org.icpc.tools.cds.video.VideoStream.StreamType;
 import org.icpc.tools.contest.Trace;
 import org.icpc.tools.contest.model.ContestUtil;
 import org.icpc.tools.contest.model.IAccount;
+import org.icpc.tools.contest.model.IAward;
 import org.icpc.tools.contest.model.IContest;
 import org.icpc.tools.contest.model.IContestObject;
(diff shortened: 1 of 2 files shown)


CDS/src/org/icpc/tools/cds/ConfiguredContest.java
```diff
@@ -983,6 +995,7 @@ public synchronized ResolutionControl initResolution(boolean startFromJudgeQueue
 		ResolverLogic resolver = new ResolverLogic(contest, false);
 		List<ResolutionStep> steps = resolver.resolveFrom(startFromJudgeQueue);
 		resolutionControl = new ResolutionControl(steps);
+		resolutionControl.setSpeedFactor(0.2);
```
nickygerritsen: Is this intended?

nickygerritsen (approved):

CDS/src/org/icpc/tools/cds/ConfiguredContest.java
deboer-tim: No.

Woke up this morning thinking 'wait, did I remember to remove the line I added to test faster?' 😂

Good catch, thank you! Line removed in new commit.

